PR #9, opened 2024-02-26
Make path to ORFS configurable

lpawelcz:
This PR moves the task of specifying path to ORFS outside of this repository and build systems that are used here.
It hardcodes an implicit dependency on `orfs_path.sh` script which is expected to be placed in the home directory. The script must contain a definition of `ORFS` environment variable with a path to local installation of `OpenROAD-flow-script`.

This modification breaks bazel's [hermeticity](https://bazel.build/basics/hermeticity) in order to allow changing the path to local ORFS installation without the need of rebuilding expensive targets.

Additionally, it is important to remember that changing version of ORFS, OpenROAD, yosys, etc., as well as any manual modification to the build system, TCL scripts, environment variables configuration and tools source code in the scope of ORFS repository **will not** cause invalidation of previous builds. It is the user's responsibility to rebuild artifacts when necessary to keep the integrity of the build after modifications made to local OpenROAD-flow-scripts setup.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
     name: Build sample stage targets
     runs-on: ubuntu-22.04
     container:
-      image: ghcr.io/antmicro/bazel-orfs@sha256:78b4c15830d75e026dc2294b280cb5ff1200f6ee7902a714dca71fd61b5b2e4e
+      image: ghcr.io/antmicro/bazel-orfs@sha256:555bd997a785b15a694687739a027b74ac4c3bdaee3bed1fb631ed48241c1b6f
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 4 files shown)


oharboe (changes requested):
This is just kicking the can down the road.

We need a more satisfactory solution.

I think the answer is to stop fighting Bazel and go for a more idiomatic solution.

1. add a docker url and version string in BUILD.bazel and use this in the build
2. for local hacking, rely exclusively on the _make scripts.

oharboe:
This is an ./orfs equivalent example for building w docker. Copy and adapt here: https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/blob/master/flow/util/docker_shell

lpawelcz:
Let's close this PR and focus on providing a solution that includes defining ORFS as an explicit Bazel dependency

PR #18, opened 2024-03-22
cts: include ADDITIONAL_LIBS in variables to CTS stage

oharboe:
Ideally we provide only the variables relevant for a stage to that stage. Is ADDITIONAL_LIBS relevant, other than for `make cts_issue`?

What about ADDITIONAL_LEFS?

This is not strictly necessary for the execution of this stage, but it is useful, otherwise .lib files for macros are not included in make cts_issue.

### DIFF
--- a/openroad.bzl
+++ b/openroad.bzl
@@ -178,6 +178,7 @@ def build_openroad(
 
     x = map(lambda ext: map2(lambda m: "//:results/" + platform + "/%s/%s/%s.%s" % (m, macro_variants.get(m, macro_variant), m, ext), macros), ["lef", "lib"])
     macro_lef_targets, macro_lib_targets = x
+    # macro_gds_targets = map(lambda m: "//:results/" + platform + "/%s/%s/6_final.gds" % (m, macro_variants.get(m, macro_variant)), macros)
 
     stage_sources = dict(stage_sources)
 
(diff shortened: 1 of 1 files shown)


oharboe:
@lpawelcz We'll need a `bazelisk build *_issue` target that creates a `_make` file which includes all variables for all stages. This will be used for make issue use-case.

maliberty:
> This is not strictly necessary for the execution of this stage,

If your cts stage matches ORFS then the .lib are necessary for repair_timing.

oharboe:
> > This is not strictly necessary for the execution of this stage,
> 
> If your cts stage matches ORFS then the .lib are necessary for repair_timing.

Why dont I get an error without the .lib files in CTS?

maliberty:
sta will treat them as black boxes with no timing (as will proprietary tools).

oharboe:
> sta will treat them as black boxes with no timing (as will proprietary tools).

how can I tell which stages use .lef and .def files?

maliberty:
lef/def are needed just at floorplan and gds generation as they are stored in odb.  Liberty are not and are needed at any stage that does timing - it is simplest to assume all stages.

oharboe:
> lef/def are needed just at floorplan and gds generation as they are stored in odb. Liberty are not and are needed at any stage that does timing - it is simplest to assume all stages.

Fixed.

PR #22, opened 2024-04-14
docker: default docker image name is same as ORFS

oharboe:
the user can specify a locked down version themself for production.

For local testing, the same docker tag as ORFS is
convenient as that makes compiling a docker image
from ORFS the regular way an obvious alternative
to using "bazel run orfs_env"

This begs the question as to why "bazel run orfs_env" now exists... Isn't that a CI concern for regression testing of bazel-orfs and not a user-concern?

Users, such as megaboom, are responsible for publishing and specifying the docker tag and url or furnishing the docker image in some out of bounds way(such as compiling locally).

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -155,7 +155,7 @@ build_openroad(
 oci_tarball(
     name = "orfs_env",
     image = "@orfs_image",
-    repo_tags = ["bazel-orfs/orfs_env:latest"],
+    repo_tags = ["openroad/flow-ubuntu22.04-builder:latest"],
 )
 
 sh_binary(
--- a/openroad.bzl
+++ b/openroad.bzl
@@ -497,7 +497,7 @@ def build_openroad(
         mock_area = None,
         platform = "asap7",
         macro_variant = "base",
-        docker_image = "bazel-orfs/orfs_env:latest"):
+        docker_image = "openroad/flow-ubuntu22.04-builder:latest"):
     """
     Spawns targets for running physical design flow with OpenROAD-flow-scripts.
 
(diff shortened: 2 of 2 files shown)


oharboe:
@lpawelcz Thoughts?

lpawelcz:
Yes, It would make sense to use the same name as ORFS defaults to. However, it needs more work on getting CI to work.
`bazel run orfs_env` can be potentially removed but then it would be up to the user to pull or provide in other way the right image.
Also, having the upstream ORFS to build and push the docker images to official container registry would be for sure a great improvement here.

lpawelcz:
@oharboe do you want us to take the code from this PR and extend it with the following changes?

* fixes to the CI
* removal of the `bazel run orfs_env` - We will set the default to `ghcr.io/antmicro/openroad-flow-scripts/ubuntu22.04:latest` which will cause docker to automatically pull this image if it is not available locally. In case of using `docker_image` attribute to override the docker image, user must provide the image himself.

side note: we published updated version of the ORFS image: https://github.com/antmicro/OpenROAD-flow-scripts/pkgs/container/openroad-flow-scripts%2Fubuntu22.04/203722065?tag=latest
It was built on top of  https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/4ac2d011db0a3516f79ee8910eb5fb74e663ef74. 
It includes the changes you referenced in some of the PRs and issues:
* https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1912
* https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1936

oharboe:
@lpawelcz Yes; please extend this PR so that it covers all CI issues anything else that was broken by the change...

Can we change the CI so that it builds the latest OpenROAD-flow-scripts docker image?

If bazel-orfs does not work with the latest ORFS, then it has to be fixed, or this project is dead, so no point in locking down ORFS for testing  bazel-orfs.

megaboom, on the other hand, should specify a specific version of ORFS that it works with.

lpawelcz:
@oharboe 

It turned out that `orfs_env` target is still useful. In fact, without it it wouldn't be possible to specify `openroad/flow-ubuntu22.04-builder:latest` as the default docker image name.

We could set the default image to `ghcr.io/antmicro/openroad-flow-scripts/ubuntu22.04:latest` and that would allow us to remove `orfs_env` target and `oci_rules` dependency in bazel workspace because first docker flow job would pull the image directly from `ghcr.io/antmicro/openroad-flow-scripts/ubuntu22.04:latest` if it was not available locally.

On the other hand, this name is not compatible with the default name of the docker image built locally with ORFS sources (`openroad/flow-ubuntu22.04-builder:latest`).

I think it would be good to have this compatibility - this was your initial idea for this PR.
With obtaining the docker image through `bazel run orfs_env` we can specify the full URL to the image and so that bazel will download it and load to local docker runtime under compatible name `openroad/flow-ubuntu22.04-builder:latest`

I don't have the right permissions to push commits to this branch so I've opened another PR (https://github.com/The-OpenROAD-Project/bazel-orfs/pull/25) to supersede this one.

PR #24, opened 2024-04-16
Fix permission denied errors

eszpotanski:
This PR addresses the permission issues, which results in failing with `Permission error` or getting stuck with `[...] overriding mode 0555 (r-xr-xr-x)?`. It makes sure that files in workspace (`bazel-bin` directory) are writable and can be overridden.

### DIFF
--- a/docker_shell.sh
+++ b/docker_shell.sh
@@ -44,6 +44,9 @@ fi
 export DESIGN_CONFIG_PREFIXED=$WORKSPACE_EXECROOT/$DESIGN_CONFIG
 export STAGE_CONFIG_PREFIXED=$WORKSPACE_EXECROOT/$STAGE_CONFIG
 
+# Make bazel-bin writable
+chmod -R +w $WORKSPACE_EXECROOT/bazel-out/k8-fastbuild/bin
+
 # Most of these options below has to do with allowing to
 # run the OpenROAD GUI from within Docker.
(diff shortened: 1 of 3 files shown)


docker_shell.sh
```diff
@@ -44,6 +44,9 @@ fi
 export DESIGN_CONFIG_PREFIXED=$WORKSPACE_EXECROOT/$DESIGN_CONFIG
 export STAGE_CONFIG_PREFIXED=$WORKSPACE_EXECROOT/$STAGE_CONFIG
 
+# Make bazel-bin writable
+chmod -R +w $WORKSPACE_EXECROOT/bazel-out/k8-fastbuild/bin
```
oharboe, line 48: It is a bit unnerving that files are overwritten in this folder. I something begin rebuilt that shouldn't?

orfs
```diff
@@ -15,6 +15,9 @@ fi
 
 export WORK_HOME=$BUILD_DIR/$RULEDIR
 
+# Make bazel-bin writable
+chmod -R +w $WORK_HOME
```
oharboe, line 19: Hmm... OK for now, but it seems to me that this should be in the _make scripts. Are we hiding a more subtle problem?

oharboe:
@lpawelcz I am a little bit unnerved that we are hiding a real problem and making it harder to find later... Thoughts?

PR #33, opened 2024-04-24
DO NOT MERGE purge memory targets, now built into ORFS

oharboe:
@lpawelcz Can you fix this PR so that it purges the memory targets? memory targets are now built into ORFS, so it is no longer a bazel-orfs cocern.


lpawelcz:
Sure, we will handle that. I imagine that you would like to be able to call `memory` make targets from ORFS with scripts generated in `bazel-orfs`, is that right?
Looking at https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1727 we notice that the memory report is now also a part of the synthesis so it should be sufficient to just test `memory` make targets from ORFS with scripts generated in `bazel-orfs`

oharboe:
> Sure, we will handle that. I imagine that you would like to be able to call `memory` make targets from ORFS with scripts generated in `bazel-orfs`, is that right? Looking at [The-OpenROAD-Project/OpenROAD-flow-scripts#1727](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/pull/1727) we notice that the memory report is now also a part of the synthesis so it should be sufficient to just test `memory` make targets from ORFS with scripts generated in `bazel-orfs`

Yes. Though you don't need to add a CI test for this. This is now just one of many, many ORFS features that just works out of the box and that we don't need CI bazel-orfs tests for.

oharboe:
this was just an example, easier to do from scratch when other changes are in.

PR #27, opened 2024-04-17
Add targets running OpenROAD GUI with Docker image

eszpotanski:
This PR adds `*_gui` targets for Docker flow, which has to be called with `run`, e.g.:
```
bazel run L1MetadataArray_full_final_gui
```

GUI for local flow works fine, apart from synthesis stage, which requires additional LEF files - new target `*_gui_synth_make` resolves this issue.

PR #38, opened 2024-04-29
Generate scripts for local and Docker flow

eszpotanski:
This PR adds `_scripts` target, which generates two scripts, for local flow (using `orfs`) and Docker flow (using `docker_shell.sh`).

To follow https://github.com/The-OpenROAD-Project/bazel-orfs/pull/27#issuecomment-2069770213, scripts are generated with `_scripts` target, and `_make` targets depends on `_scripts` apart from building all required dependencies for the stage.

Moreover, `docker_shell` can detect whether bazel-orfs directory has to be mounted and automatically adds necessary argument to `docker run`. Therefore, no additional changes will be required, when bazel-orfs is used with `local_path_override`.

### DIFF
--- a/.github/scripts/build_local_target.sh
+++ b/.github/scripts/build_local_target.sh
@@ -3,6 +3,7 @@
 set -e
 
 target_name=${TARGET:-"tag_array_64x184"}
+flow=${FLOW:-"local_make"}
 if [[ -z "$STAGES" ]]; then
   if [[ "$target_name" == L1MetadataArray_* ]]; then
     STAGES=("synth_sdc" "synth" "floorplan" "place" "cts" "grt" "generate_abstract")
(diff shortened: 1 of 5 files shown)


oharboe (changes requested):
Nice.

I want to clean up the user experience a bit:

- `bazel build L1MetadataArray_test_generate_abstract` no longer exists, it has been renamed to `bazel build L1MetadataArray_test_generate_abstract_build`. Remove the `_build` suffix to restore the `bazel build L1MetadataArray_test_generate_abstract` target name.
- Rename `bazel-bin/tag_array_64x184_floorplan_local_script` to `bazel-bin/tag_array_64x184_floorplan_local_make`
- Rename `bazel-bin/tag_array_64x184_floorplan_docker_script` to `bazel-bin/tag_array_64x184_floorplan_docker`

eszpotanski:
I renamed targets according to the comment

PR #48, opened 2024-05-10
ORFS: Bump docker image

lpawelcz:
Use ORFS docker image built on top of https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/commit/9b86e54c852ec4e77a94b7ba3c3bfcf012d6adba


lpawelcz:
Noticed that this will fail in multiple test cases with:
```
[ERROR RSZ-0060] Max buffer count reached.
```

CC @oharboe

PR #51, opened 2024-05-14
Subpackage targets

hovind:
Trying to call `build_openroad()` from subpackages does not currently work. I've drafted a sketch to make it work when called from subpackages of `bazel-orfs`.

It still does not work when `build_openroad()` is called from a subpackage in an external project, with the following error message from `docker_shell.sh`, I believe it has to do with how `WORKSPACE_ROOT` is defined.

```
+ chmod -R +w /home/.../execroot/_main/execroot/_main/bazel-out/k8-fastbuild/bin
chmod: cannot access '/home/.../execroot/_main/execroot/_main/bazel-out/k8-fastbuild/bin
```

### DIFF
--- a/openroad.bzl
+++ b/openroad.bzl
@@ -377,8 +377,8 @@ def mock_area_stages(
             stage_args = mock_area_stage_args[stage] + ["FLOW_VARIANT=mock_area"],
         )
         make_pattern = Label("//:" + stage + "-bazel.mk")
-        design_config = Label("@@//:" + name + "_mock_area_config.mk")
-        stage_config = Label("@@//:" + name + "_" + stage + "_mock_area_config.mk")
+        design_config = Label("@@//" + native.package_name() + ":" + name + "_mock_area_config.mk")
+        stage_config = Label("@@//" + native.package_name() + ":" + name + "_" + stage + "_mock_area_config.mk")
         make_targets = get_make_targets(stage, True, mock_area)
 
         native.genrule(
@@ -512,6 +512,11 @@ def init_output_dict(all_stages, platform, out_dir, variant, name):
 
     return outs
 
+def resolve_path(label):
+    if native.package_name():
+        return "/".join([native.package_name(), label])
+    return label
+
 def build_openroad(
         name,
         variant = "base",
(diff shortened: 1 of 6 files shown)


openroad.bzl
```diff
@@ -592,7 +592,7 @@ def build_openroad(
     stage_args = init_stage_dict(all_stage_names, stage_args)
     stage_args["clock_period"] = SDC_FILE
     stage_args["synth_sdc"] = SDC_FILE
-    stage_args["synth"].append("VERILOG_FILES=" + " ".join(verilog_files))
+    stage_args["synth"].append("VERILOG_FILES=" + " ".join(map(lambda v: native.package_name() + "/" + v, verilog_files)))
```
hovind, line 595: I believe this is too simple if we want to pass in labels that are outputs of other rules.

oharboe:
@lpawelcz @eszpotanski Merging for now, but this needs a better solution.

Thoughts?

lpawelcz:
@hovind thanks for adding CI test case!

lpawelcz:
> It still does not work when build_openroad() is called from a subpackage in an external project, with the following error message from docker_shell.sh, I believe it has to do with how WORKSPACE_ROOT is defined.

I was able to get this to work in external workspace (megaboom). Please take a look at: https://github.com/The-OpenROAD-Project/megaboom/pull/59

hovind:
> > It still does not work when build_openroad() is called from a subpackage in an external project, with the following error message from docker_shell.sh, I believe it has to do with how WORKSPACE_ROOT is defined.
> 
> I was able to get this to work in external workspace (megaboom). Please take a look at: [The-OpenROAD-Project/megaboom#59](https://github.com/The-OpenROAD-Project/megaboom/pull/59)

Thanks! Not at my computer right now, but I will test first thing in the morning!

Did you try with a `local_path_override()` of `bazel-orfs`? That might have been my problem.

lpawelcz:
Yes, I also tested setup with local `bazel-orfs` and it did work

hovind:
> Yes, I also tested setup with local `bazel-orfs` and it did work

Thanks, pinning bazel to `7.0.0` instead of using `7.1.2` works!

PR #67, opened 2024-06-06
openroad: Mark executable scripts accordingly

oharboe:


oharboe:
@hovind @lpawelcz @eszpotanski Thoughts?

eszpotanski:
`executable = True` makes targets runable, so they can be called with `bazel run {target}`, which is not intended I believe.
I guess it can be a little bit confusing

PR #68, opened 2024-06-06
openroad: Resolve paths

oharboe:
@hovind @eszpotanski @lpawelcz Thoughts?

### DIFF
--- a/openroad.bzl
+++ b/openroad.bzl
@@ -317,8 +317,6 @@ def mock_area_stages(
         name,
         design_name,
         stage_sources,
-        io_constraints,
-        sdc_constraints,
         stage_args,
         outs,
         variant,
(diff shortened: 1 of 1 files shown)


openroad.bzl
```diff
@@ -828,20 +817,15 @@ def build_openroad(
...
-        stage_cfg_srcs = []
-        if sdc_constraints != None:
-            stage_cfg_srcs.append(sdc_constraints)
-        if io_constraints != None:
-            stage_cfg_srcs.append(io_constraints)
         write_stage_config(
             name = target_name + "_" + stage + "_config",
             stage = stage,
-            srcs = stage_cfg_srcs,
+            srcs = stage_sources[stage],
```
eszpotanski, line 828: Adding all stage's sources to `srcs` of config targets, changes how `_scripts` target behave. Apart from, creating scripts it will try to generate dependencies which are not required for configs.

This is also the reason why the CI fails, as it tries to build additional dependencies with the Docker flow, where only the local flow should be tested

PR #69, opened 2024-06-06
Support subpackages

oharboe:
@hovind @eszpotanski @lpawelcz Thoughts?

PR #70, opened 2024-06-10
Cross package dependencies

hovind:
@eszpotanski thanks for having a look at https://github.com/The-OpenROAD-Project/bazel-orfs/pull/68, https://github.com/The-OpenROAD-Project/bazel-orfs/pull/69.

I hacked together these commits to solve two problems I had when trying to use `bazel-orfs`.

1. Using `verilog_files` from other packages.
2. Using `macros` from other packages.

In order to easily get path to files from cross-package labels, I tried to use `$(location <label>)` expansion. Using this, any path that is to be included in a `config.mk` target, will be a dependency of the `config.mk` target.

I have made an alternative implementation that does not rely on `$(location <label>)` expansion here, but I wonder if the `$(location <label>)` expansion version isn't easier?

I've tried to clean up the hacks in https://github.com/The-OpenROAD-Project/bazel-orfs/pull/68, https://github.com/The-OpenROAD-Project/bazel-orfs/pull/69 for upstream in this pull request.


BUILD
```diff
@@ -44,6 +46,7 @@ filegroup(
         "test/io.tcl",
         ":util",
     ],
+    visibility = [":__subpackages__"],
```
hovind, line 49: Added `visibility` to be able to depend on macros from other packages.

openroad.bzl
```diff
@@ -49,22 +49,6 @@ def map(func, iterable):
         result.append(func(item))
     return result
 
-def map2(func, iterable):
```
hovind, line 52: Why does this exist?

openroad.bzl
```diff
@@ -709,15 +695,17 @@ def build_openroad(
                             ["GDS_ALLOW_EMPTY=(" + "|".join(macros) + ")"] if len(macros) > 0 else [])
     stage_args["generate_abstract"] += ["ABSTRACT_SOURCE=" + abstract_source] if mock_abstract else []
 
-    stage_sources = init_stage_dict(all_stage_names, stage_sources)
+    stage_cfg_sources = init_stage_dict(all_stage_names, {})
     if sdc_constraints != None:
-        stage_sources["synth_sdc"] = [sdc_constraints]
-        stage_sources["clock_period"] = [sdc_constraints]
-    stage_sources["synth"] = list(filter(stage_sources["synth"], lambda s: not s.endswith(".sdc")))
```
hovind, line 716: What does this filtering check achieve? Seems very fickle to me, e.g. if a `filegroup` is referenced.

openroad.bzl
```diff
@@ -891,7 +874,7 @@ def build_openroad(
                 write_stage_config(
                     name = target_name + "_gui_" + stage + "_config",
                     stage = stage,
-                    srcs = stage_cfg_srcs,
+                    srcs = stage_cfg_sources["synth"] + stage_cfg_sources["synth_sdc"],
```
hovind, line 877: `stage_args["synth_sdc"]` has a `$(location <label>)` expansion of `stage_cfg_sources["synth_sdc"]`. `lefs_args` refers to `ADDITIONAL_LEFS`, but as this is not `$(location <label>)` expanded, we do not need to encode the dependency. Should these not be treated the same way?

subpackage/BUILD
```diff
@@ -1,16 +1,23 @@
 load("//:openroad.bzl", "build_openroad")
 
 build_openroad(
-    name = "tag_array_64x184",
-    io_constraints = "//:io-sram",
-    abstract_stage = "floorplan",
-    sdc_constraints = "//:constraints-sram",
+    name = "L1MetadataArray",
+    io_constraints = "//:io",
+    macros = ["//:tag_array_64x184"],
```
hovind, line 6: Introduced an example of a `build_openroad()` call referring to a macro from a different package.

eszpotanski (commented):
You can run `buildifier -lint warn -r .` to format Bazel files and add docstrings for `visibility` param to fix [Lint Bazel files CI](https://github.com/The-OpenROAD-Project/bazel-orfs/actions/runs/9446436035/job/26016123557?pr=70)

openroad.bzl
hovind: But if `sdc_constraints` is a `filegroup()` like in the `constraints-sram` case (from `BUILD` in this repo), how does the filter achieve anything?

hovind: And these dependencies being filtered are from the user supplied `stage_sources`, should we alter those?

openroad.bzl
hovind: Yes, thanks. However it is unclear to me why we should use `$(location <label>)` for some paths, and `$(BINDIR)` for others?

openroad.bzl
eszpotanski: Yes, you're right, it probably a leftover and does nothing right now.

openroad.bzl
eszpotanski: I don't think there is a strict rule, `$(BINDIR)` gives more freedom and `$(location <label>)` should be more robust (especially for targets from external repositories). In case on `ADDITIONAL_LES/LIBS` we can use both, I believe. You can change it to `location`, it will probably be cleaner and more consistent.

hovind:
> You can run `buildifier -lint warn -r .` to format Bazel files and add docstrings for `visibility` param to fix [Lint Bazel files CI](https://github.com/The-OpenROAD-Project/bazel-orfs/actions/runs/9446436035/job/26016123557?pr=70)

Thanks, fixed the linter issues. Is there anything else you think should be fixed?

PR #72, opened 2024-06-13
Remove logs and report files from targets' inputs

eszpotanski:
This PR adds filtering of targets' inputs to not depend on log and report files.

### DIFF
--- a/openroad.bzl
+++ b/openroad.bzl
@@ -847,7 +847,7 @@ def build_openroad(
                 "//conditions:default": [Label("//:docker_shell")],
             }),
             srcs = [make_pattern, design_config, stage_config] + stage_sources[stage] +
-                   ([target_name + "_" + previous] if stage not in ("clock_period", "synth_sdc") else []) +
+                   ([i for i in outs[previous] if len(i) > 4 and i[-4:] != ".log" and i[-4:] != ".rpt"] if stage not in ("clock_period", "synth_sdc") else []) +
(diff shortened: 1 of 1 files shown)


openroad.bzl
```diff
@@ -297,6 +297,19 @@ def get_entrypoint_cmd(
...
+    Filters out log and report files from stage's output.
+
+    Args:
+      outs: dictionary of lists with paths to output files for each flow stage
+      stage: name of the stage
+
+    Retruns:
+      list with filtered output files
+    """
+    return [i for i in outs.get(stage, []) if len(i) > 4 and i[-4:] not in (".log", ".rpt")]
```
hovind, line 311: Should we use https://bazel.build/rules/lib/core/string#endswith?

> sub 	[string](https://bazel.build/rules/lib/core/string.html); or [tuple](https://bazel.build/rules/lib/core/tuple.html) of [string](https://bazel.build/rules/lib/core/string.html)s; required
The suffix (or tuple of alternative suffixes) to match.

hovind (approved):
Seems reasonable to me, nice!

openroad.bzl
oharboe: endswith doesn't jell with "in (...)", so it would have to be map... Good 'nuf for now.

oharboe:
@eszpotanski Did a squash and merge this time. I suppose it was the intention to squash these two commits, but it wasn't done for some reason?

PR #74, opened 2024-06-26
openroad: Support generated verilog

hovind:
The old approach doesn't work for generated verilog files. @eszpotanski does this seem reasonable?


eszpotanski (approved):
LGTM

PR #75, opened 2024-07-01
Dockerless orfs

hovind:
A few things with the current implementation have been causing me trouble.

1. Some artifacts generated by the current rules are not meant to be executed within `bazel`, making it difficult to e.g. integrate a report artifact into the `bazel` build.
2. The current implementation does not work for `bazel` version `7.2.0`, it relies on a lot of implementation details for things like artifact locations.
3. We need to use `--spawn_strategy=local` to build using docker, allowing `bazel` to be blissfully unaware of some dependencies
4. Everything is implemented in term of `bazel` macros (`def`s), instead of `rule`s. This makes things quite a bit harder to implement than necessary (imo), as we can't use the more expressive `ctx` API.

This is a feature incomplete draft meant to garner feedback, and I am curious what you think about the approach @eszpotanski. Its implementation comes with its own set of hacks, hopefully we can improve it over time if we decide to go with this approach.


docker.BUILD.bazel
```diff
@@ -0,0 +1,472 @@
...
+    shared_library = ":libortools.so.9",
+)
+
+cc_import(
+    name = "libtclreadline",
+    shared_library = ":libtclreadline-2.3.8.so",
+)
+
+cc_import_binary(
+    name = "openroad_raw",
```
hovind, line 72: This is quite a lot of boilerplate to import the binaries from the docker image, we can probably make a macro to make it less verbose.

There is also the option of using the binaries built in https://github.com/hdl/bazel_rules_hdl/, but then we are no longer guaranteed to have `openroad`, `yosys`, `Makefile` etc. from the exact same https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts commit.

docker.BUILD.bazel
```diff
@@ -0,0 +1,472 @@
...
+)
+
+stringpatch_binary(
+    name = "openroad",
+    data = [
+        ":tcl/tclreadlineInit.tcl",
+    ],
+    src = ":openroad_raw",
+    replacements = {
+        "/usr/lib/tcltk/x86_64-linux-gnu/tclreadline2.3.8/tclreadlineInit.tcl": "$(location :tcl/tclreadlineInit.tcl)",
```
hovind, line 469: Another hard coded path.

BUILD
```diff
@@ -53,171 +59,161 @@ filegroup(
...
-        ],
-    },
-    variant = "external_pdk",
-    verilog_files = ["test/rtl/L1MetadataArray.sv"],
+abstract(
+    name = "array_abstract",
+    src = ":array_place",
+)
+
+orfs_open(
```
hovind, line 201: Added an example for the `orfs_open` rule.

docker.bzl
```diff
@@ -0,0 +1,65 @@
...
+        repository_ctx.extract(archive = image_archive)
+
+    repository_ctx.symlink(repository_ctx.attr.build_file, "BUILD")
+    patch(repository_ctx)
+
+docker_pkg = repository_rule(
+    implementation = _impl,
+    attrs = {
+        "build_file": attr.label(mandatory = True),
+        "image": attr.string(mandatory = True),
```
hovind, line 49: Changed to refer to `"image"` instead of `"docker_file"` now that `orfs` images are uploaded regularly.

BUILD
```diff
@@ -53,171 +59,147 @@ filegroup(
...
+    module_top = "tag_array_64x184",
+    verilog_files = ["test/mock/tag_array_64x184.sv"],
+)
+
+SRAM_FLOOR_PLACE_ARGUMENTS = {
+    "IO_CONSTRAINTS": "$(location :io-sram)",
+}
+
+floorplan(
+    name = "tag_array_floor",
```
oharboe, line 92: I don't like that we have to invent names here for each stage. The user now has to come up with lots and lots of names and even within this example, it is done inconsistently.

I don't have a suggestion on how to address this.

BUILD
```diff
@@ -53,171 +59,147 @@ filegroup(
...
+        "SDC_FILE": "$(location :constraints-sram)",
+    },
+    data = [
+        ":constraints-sram",
+    ],
+    module_top = "lb_32x128",
     verilog_files = ["test/rtl/lb_32x128.sv"],
 )
 
-build_openroad(
```
oharboe, line 100: Can we still support the build_openroad() macro and have it expand to these new rules?

docker.BUILD.bazel
```diff
@@ -0,0 +1,472 @@
...
+    visibility = ["//visibility:public"],
+)
+
+pdk(
+    name = "asap7",
+    srcs = glob([
+        "flow/platforms/asap7/**/*.gds",
+        "flow/platforms/asap7/**/*.lib.gz",
+        "flow/platforms/asap7/**/*.lef",
+        "flow/platforms/asap7/**/*.lib",
```
oharboe, line 39: This won't work for all the other PDKs. The nice thing about using a docker image is that we don't have all this whitebox stuff bleed into bazel-orfs...

oharboe (changes requested):
Some feedback...

oharboe:
@hovind All the CI tests are broken...

hovind:
> @hovind All the CI tests are broken...

> This is a feature incomplete draft meant to garner feedback, and I am curious what you think about the approach @eszpotanski. Its implementation comes with its own set of hacks, hopefully we can improve it over time if we decide to go with this approach.

BUILD
hovind: I agree that it's verbose, but how are you supposed to refer to the output of a stage without giving it a name?

BUILD
oharboe: Make a naming convention through the build_openroad() macro?

docker.BUILD.bazel
```diff
@@ -0,0 +1,483 @@
...
+        "flow/platforms/asap7/**/*.rules",
+        "flow/platforms/asap7/**/*.sdc",
+        "flow/platforms/asap7/**/*.tcl",
+        "flow/platforms/asap7/**/*.v",
+        "flow/platforms/common/**/*.v",
+    ]),
+    visibility = ["//visibility:public"],
+)
+
+cc_import_library(
```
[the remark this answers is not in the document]
eszpotanski: Is it not enough to use ORFS image with pinned version?

BUILD
hovind: Yes we can!

eszpotanski (commented):
About the points:
1. Yes, right now caching only works with Stage targets, results generated through scripts from `bazel-bin` won't be cached.
2. Does Bazel version 7.2.0 has some feature we need?
3. Using Docker flow, most dependencies are defined by Docker image, the other ones are design files, so both are available to Bazel and if one of those change, results will be rebuilt, or am I missing something?
4. Sure, rules with `ctx` API are more verbose and human-readable, but on the other side they require more boilerplate, so usually they were used when some custom behavior had to be introduced, like for [cpu_heavy_genrule](https://github.com/The-OpenROAD-Project/bazel-orfs/blob/main/openroad.bzl#L606). Rewriting them now will probably introduce some regressions, but it can be done.

hovind:
> About the points:
> 
>     1. Yes, right now caching only works with Stage targets, results generated through scripts from `bazel-bin` won't be cached.
> 
>     2. Does Bazel version 7.2.0 has some feature we need?
> 
>     3. Using Docker flow, most dependencies are defined by Docker image, the other ones are design files, so both are available to Bazel and if one of those change, results will be rebuilt, or am I missing something?
> 
>     4. Sure, rules with `ctx` API are more verbose and human-readable, but on the other side they require more boilerplate, so usually they were used when some custom behavior had to be introduced, like for [cpu_heavy_genrule](https://github.com/The-OpenROAD-Project/bazel-orfs/blob/main/openroad.bzl#L606). Rewriting them now will probably introduce some regressions, but it can be done.

Thanks for the feedback!

2. Not right now, but I think it's better not to depend on implementation details of a certain `bazel` version if we can avoid it.
3. What I was trying to express is that `bazel` no longer verifies dependencies for us with the `--spawn-strategy=local` flag. We can read any file on the filesystem without having declared it in a `genrule`'s `srcs`.
4. I see. The rules approach also allows us to use https://bazel.build/reference/be/functions#select. Using `select()` with macros only works if the selected value is passed directly to a rule attribute, which is rarely the case.

Makefile.patch
```diff
@@ -0,0 +1,16 @@
...
+@@ -1,3 +1,4 @@
++THIS_FILE = $(MAKEFILE_LIST)
+ # settings.mk is not under source control. Put variables into this
+ # file to avoid having to adding the to the make command line.
+ -include settings.mk
+@@ -549,7 +550,7 @@
+ 
+ UNSET_VARS = for var in $(UNSET_VARIABLES_NAMES); do unset $$var; done
+ 
+-SUB_MAKE = $(MAKE) $(foreach V,$(COMMAND_LINE_ARGS), $(if $($V),$V=$(shell echo "$($V)" | $(FLOW_HOME)/scripts/escape.sh),$V='')) --no-print-directory DESIGN_CONFIG=$(DESIGN_CONFIG)
```
oharboe, line 12: upstream

openroad.bzl
```diff
@@ -540,12 +540,8 @@ place = rule(
             "3_place.odb",
             "3_place.sdc",
         ],
-        log_names = [
-            "3_1_place_gp_skip_io.log",
-        ],
-        report_names = [
-            "5_global_place.rpt",
-        ],
+        log_names = [],
```
oharboe, line 543: These should be fixed by adding a "touch" to bazel-place.mk

oharboe (changes requested):
tweaks

PR #77, opened 2024-07-03
Make `docker_image` required

eszpotanski:
This PR removed the default value of `docker_image` argument and forces to specify Docker image and its version for `build_openroad()` macro.

Also, CI was changed to use official `openroad/orfs` image.


oharboe:
@hovind Please incorporate into your rework of build_openroad() as relevant.

PR #90, opened 2024-08-06
orfs_flow: test SKIP_REPORT_METRICS=1

oharboe:
This requires creating an empty file when reports are disabled because bazel does not have conditional outputs


oharboe:
@ivysochyn If this works, I'll merge it. It could do with some polish, but the important part is the concept and how responsibilities are divided between bazel and ORFS.

PR #92, opened 2024-08-07
 Avoid copying source files to the output

ivysochyn:
# Avoid copying source files to the output directory

This PR addresses the unnecessary copying of source files to the output directory in the `openroad.bzl` script.

## Example

Before modifications:

```
$ bazel run L1MetadataArray_place -- `pwd`/build
$ find build/ | grep -v external | grep floorplan
build/results/asap7/L1MetadataArray/base/2_floorplan.odb
build/results/asap7/L1MetadataArray/base/2_floorplan.sdc
```

After:

```
$ bazel run L1MetadataArray_place -- `pwd`/build
$ find build/ | grep -v external | grep floorplan
```

## Modifications

This PR modifies the following:

* `openroad.bzl`
  * Modified the `GENFILES` substitution to exclude `ctx.files.src`, ensuring that source files are not copied to the output directory.
  * Updated the runfiles section to exclude `ctx.files.src`, aligning it with the updated `GENFILES` logic.

### DIFF
--- a/openroad.bzl
+++ b/openroad.bzl
@@ -625,7 +625,7 @@ def _make_impl(ctx, stage, steps, result_names = [], object_names = [], log_name
         template = ctx.file._deploy_template,
         output = exe,
         substitutions = {
-            "${GENFILES}": " ".join([f.short_path for f in [config_short] + results + ctx.files.src + ctx.files.data]),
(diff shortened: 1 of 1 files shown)


oharboe:
@hovind FYI

PR #95, opened 2024-08-08
MODULE.bazel: don't keep MODULE.bazel.lock under version control

oharboe:
@ivysochyn Thoughts?


ivysochyn:
I believe we should version control the `MODULE.bazel.lock` file. Doing so ensures reproducibility, consistency, and traceability of our builds, making it easier to manage dependencies. The Bazel documentation outlines these benefits clearly [here](https://bazel.build/external/lockfile#lockfile-benefits). Other tools like `npm` also use similar lock files (e.g., `package-lock.json`) for the same reasons.

Overall, keeping this file in version control aligns with best practices as far as I know.

oharboe:
What is the procedure for updating MODULE.bazel then? 

1. modify MODULE.bazel
2. Run mystery command :-)
3. create commit with MODULE.bazel*
4. push

?

PR #96, opened 2024-08-09
.bazelrc: delete, it is a docker vestige

oharboe:
'build --incompatible_strict_action_env' is no longer needed


oharboe:
@eszpotanski OK?

PR #114, opened 2024-08-21
Speed up ORFS version change

eszpotanski:
This PR changes how the `@docker_orfs` is created to speed up the process of changing the ORFS version.

First of all, instead of exporting Docker image to TAR file and extracting it, Docker container is run and files are copied directly from it.

Moreover, `patcher.py` script was parallelized - read-only commands are run in parallel preparing `patchelf` commands updating linking information, which are later on run sequentially.

When updating ORFS version from `v3.0-1114-g46acc762` to `v3.0-1154-gdfcf7c64`, following speedup can be observed:
* without cached image - decrease from 4m to 2m45s,
* with cached image - decrease from 2m30s to 1m30s.


oharboe:
@hovind FYI

PR #127, opened 2024-09-03
Make GRT a separate step

mleonowicz:
This PR introduces `grt` as a separate stage from `route`.

Now, stages may have multiple source stages instead of only one. This was needed, as the new `route` stage needs both `congestion.rpt` file, which comes from the `grt` route, and `4_cts.sdc` file, which is produced by a `cts` stage. 

Attaching `before` and `after` outputs produced by:
`bazel query "deps(//:L1MetadataArray_generate_abstract)" --output=graph --noimplicit_deps --notool_deps > graph && xdot graph`

before
after

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -92,6 +92,9 @@ orfs_flow(
     },
     verilog_files = ["//another:tag_array_64x184.sv"],
     visibility = [":__subpackages__"],
+    common_args = {
+        "GENERATE_ARTIFACTS_ON_FAILURE": "1",
+    },
 )
 
 LB_STAGE_ARGS = {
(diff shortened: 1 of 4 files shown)


oharboe (changes requested):
- Set GENERATE_ARTIFACTS_ON_FAILURE=1 variable for all stages in ORFS. This will set ORFS into a mode where it creates an artifact for a failed operation, such as ORFS so that e.g. a global route failure artifacts can be downloaded and studied.
- Rebase on main

mleonowicz:
To support passing `GENERATE_ARTIFACTS_ON_FAILURE = 1` argument to all stages, I have added a new argument to `orfs_flow` called `common_args`.

I have also rebased the branch.

PR #130, opened 2024-09-04
openroad: Combine canonicalization and synth actions into one step

hovind:
@oharboe FYI.
@mleonowicz what do you think about https://github.com/The-OpenROAD-Project/bazel-orfs/commit/c9861eb4fee927d2f8029734999ae3c1fdaaffc2 instead of adding multiple sources?

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -92,9 +92,6 @@ orfs_flow(
     },
     verilog_files = ["//another:tag_array_64x184.sv"],
     visibility = [":__subpackages__"],
-    common_args = {
-        "GENERATE_ARTIFACTS_ON_FAILURE": "1",
-    },
 )
 
 LB_STAGE_ARGS = {
@@ -120,9 +117,6 @@ orfs_flow(
     stage_args = LB_STAGE_ARGS,
     stage_sources = LB_STAGE_SOURCES,
     verilog_files = LB_VERILOG_FILES,
-    common_args = {
-        "GENERATE_ARTIFACTS_ON_FAILURE": "1",
-    },
 )
 
 # buildifier: disable=duplicated-name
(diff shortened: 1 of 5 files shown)


oharboe (changes requested):
fix bazel lint

mleonowicz:
@hovind 
I think it the `forwarded_names` approach looks good as well. I decided to go with multiple sources to be more explicit about the dependency graph and to avoid having to manually choose files that are to be passed to the `grt` stage (in a similar way we do not pass `congestion.rpt` file to the `grt` stage)

hovind:
> @hovind I think it the `forwarded_names` approach looks good as well. I decided to go with multiple sources to be more explicit about the dependency graph and to avoid having to manually choose files that are to be passed to the `route` stage (in a similar way we do not pass `congestion.rpt` file to the `route` stage)

hovind:
efer 

> @hovind I think it the `forwarded_names` approach looks good as well. I decided to go with multiple sources to be more explicit about the dependency graph and to avoid having to manually choose files that are to be passed to the `route` stage (in a similar way we do not pass `congestion.rpt` file to the `route` stage)

hovind:
> @hovind I think it the `forwarded_names` approach looks good as well. I decided to go with multiple sources to be more explicit about the dependency graph and to avoid having to manually choose files that are to be passed to the `route` stage (in a similar way we do not pass `congestion.rpt` file to the `route` stage)

Great! I prefer adding a reusable mechanism (`forwarded_names`) to having to perform surgery like
```
        if step.stage == 'route':
            srcs.append(name_template.format(name, "cts"))
```
and
```
(previous_reports if stage == '5_2_route' else [])
```

PR #140, opened 2024-09-09
bazel-orfs: centralize mapping of variables onto stages

oharboe:
The user no longer has to guess which argument goes into which stage, but the list of this mapping is hardcoded into bazel-orfs.

If an argument is missing from STAGE_ARGS_USES, an error message is printed that includes:

if stage in STAGE_ARGS_USES[arg]}
Error: key "SDC_FILE" not found in dictionary


oharboe:
@hovind Thoughts?

PR #143, opened 2024-09-10
Allow using locally built images

glatosinski:
This PR introduces option to use locally build images - if the sha256 is left empty, the `docker.bzl` checks if the given `image` is present locally and uses it.


oharboe:
@jeffng-or FYI

PR #151, opened 2024-09-15
sram: fakeram vs. mock SRAM

oharboe:


oharboe:
@jeffng-or FYI, will add fake sram tomorrow and write some docs. There are some snags to be fixed in bazel-orfs as well...

oharboe:
@maliberty @jeffng-or There's no clock and "all_registers" returns 0 with fakeram. Is there something wrong with the .lib file?

Either use the standalone .tar below, or with this PR run:

```
bazel run //sram:top_fakeram_floorplan `pwd`/build
build/make gui_floorplan
```



untar [no-clock-or-registers.tar.gz](https://github.com/user-attachments/files/17006561/no-clock-or-registers.tar.gz)

```
./run-me-top-asap7-fakeram.sh
```

image

```
>>> llength [all_registers]
0
[WARNING GUI-0097] All pins are unconstrained. Cannot plot histogram. Check if timing data is loaded!
>>> report_clock_properties
Clock                   Period          Waveform
----------------------------------------------------
```

sram/README.md
```diff
@@ -0,0 +1,31 @@
...
+=====================
+
+fakeram is a .lib, .lef and .v file generated by the fakeram tool.
+
+mocked SRAM takes as input the design's behavioral model Verilog file and is used to generate a .lef and .lib file.
+
+Mock SRAM based on fakeram's behavioral .v file
+-----------------------------------------------
+
+The .v file generated by fakeram does not build with yosys unmodified. The fakeram/sdq_17x64.sv has been manually pruend until it passed yosys.
```
hovind, line 11: s/pruend/pruned

jeffng-or:
The fact that there are no clocks make sense. The sram/fakeram/constraints-sram.sdc defines the clock variables, but the call to source $PLATFORM_DIR/constraints.sdc is being gated by the all_registers check. Since all_registers is returning an empty list, we never get to create the clocks.

In terms of all_registers, the liberty only has pin definitions, so not sure what we expect all_registers to report.

oharboe:
If the .lib does not have any registers how can it store state? The symptom is no clock, the cause is that all_registers returns empty...

PR #201, opened 2024-10-11
naja: example of how to create a base and naja variant

oharboe:
@xtofalex FYI


xtofalex:
> @xtofalex FYI

So I need to apply same kind of patch to top megaboom BUILD.bazel file ?

oharboe:
> > @xtofalex FYI
> 
> So I need to apply same kind of patch to top megaboom BUILD.bazel file ?

Yes.

PR #223, opened 2024-10-17
build: rename inputs instead of outputs

oharboe:


openroad.bzl
```diff
@@ -405,9 +405,9 @@ def yosys_only_attrs():
         ),
     }
 
-def output_attrs(outputs):
+def input_attrs(outputs):
```
hovind, line 408: Nit: `outputs` could be renamed, but should probably be removed and replaced by `default = {}`?

openroad.bzl
```diff
@@ -1258,7 +1284,7 @@ def orfs_flow(
       sources: dictionary keyed by ORFS variables with lists of sources
       stage_sources: dictionary keyed by ORFS stages with lists of stage-specific sources
       stage_arguments: dictionary keyed by ORFS stages with lists of stage-specific arguments
-      stage_outputs: dictionary keyed by ORFS stages to rename outputs
+      stage_inputs: dictionary keyed by ORFS stages to rename outputs
```
hovind, line 1287: Nit: Outdated docstring.

hovind (approved):

PR #225, opened 2024-10-17
Updated sram BUILD with naja command line

xtofalex:
bazel build naja launches naja_edit.
For primitives, the embedded asap7 Python version in naja is currently used. It would be better to point to the liberty file provided to Yosys.


xtofalex:
naja package must be installed in naja/install dir inside sram.

PR #233, opened 2024-10-21
openroad.bzl: Rename deps

hovind:
@oharboe


oharboe:
Very confusing! I tried it locally and I saw it both succeed and fail... How can it be flaky???

PR #252, opened 2024-12-07
Chisel rules

oharboe:
Not particularly pretty or easy to add Chisel builds


oharboe:
out of sight out of mind until scalabin switches to bazelmod

PR #286, opened 2025-02-17
netlistsvg: example of how to generate a .svg from a .v file

oharboe:


maliberty:
Do you find this useful?  Schematics usually don't scale up well for large designs.

oharboe:
> Do you find this useful? Schematics usually don't scale up well for large designs.

It is useful to generate simple diagrams for documentation purposes but where the Verilog is written explicitly and have been greatly simplified documentation purposes.

bazel-orfs might not be the right place for this longer term, but I'm parking this code here for now. Also, I was able to sort through how to run yosys in standalone mode.


To try:

```
bazel build alu_svg
```

```verilog
module ALU(
    input clk,
    input [31:0] a,
    input [31:0] b,
    input [2:0] op,
    output reg [31:0] out
    );
    reg [31:0] aReg;
    reg [31:0] bReg;
    reg [2:0] opReg;

    always @*
    begin
        case(opReg)
            3'b001: out = aReg + bReg;
            3'b010: out = aReg & bReg;
            3'b011: out = aReg | bReg;
            default: out = 32'd0;       // Handle other cases or add a default value
        endcase
    end
    always @(posedge clk) begin
        aReg <= a;
        bReg <= b;
        opReg <= op;
    end
endmodule
```


image

oharboe:
@maliberty bazel shines here because it handles the nasty dependency on yosys as well as netlistsvg(npm) to produce a .svg from a .v file. Doing this from Sphinx is nasty. Doing this for .md files is nasty because you have to introduce some way to cope with these dependencies.

PR #303, opened 2025-03-19
orfs: v3.0-2623-g372e2136

oharboe:


oharboe:
@eder-matheus @maliberty Last known good ORFS ref https://github.com/The-OpenROAD-Project/OpenROAD/issues/6874

PR #304, opened 2025-03-20
Bump python bazel dependency and toolchain

hovind:
I don't think the `requirement()` function contributes more  than it obfuscates, but feel free to remove that commit. @oharboe


oharboe:
@hovind Linting error

PR #327, opened 2025-04-25
Yosys oss

oharboe:


oharboe:
@vvbandeira Hmm.... the build isn't queued. Is there a problem?

PR #328, opened 2025-05-08
Orfs bump

oharboe:


oharboe:
@povik FYI, added slang smoketest to bazel-orfs

PR #333, opened 2025-05-12
view: bazelisk run foo_grt now works

oharboe:
Retire code that was messing up paths.


oharboe:
superseded https://github.com/The-OpenROAD-Project/bazel-orfs/pull/335

PR #336, opened 2025-05-20
deps: fix foo_deps target for list variables

oharboe:
ADDITIONAL_LEFS did not have the paths fixed as only the first path in the array was fixed


oharboe:
@hovind FYI

PR #363, opened 2025-08-15
tests: workaround for placement snag, part 2

oharboe:
Modify initial conditions very slightly so that we don't run into https://github.com/The-OpenROAD-Project/OpenROAD/issues/7745

revert this commit when OpenROAD is fixed.


oharboe:
@maliberty bazel-orfs tests working again :relieved:

PR #380, opened 2025-09-20
sram: use BLOCK and BLOCKS PDN files from asap7

oharboe:


oharboe:
@maliberty FYI. PDN errors were mostly to do with BLOCK for macros and BLOCKS for where macros were used + increase sizes a bit to fit PDN

PR #422, opened 2025-12-14
Use Maven llvm-firtool JAR to fix GLIBC compatibility (v1.137.0)

MrAMS:
Replace circt-full-static-linux-x64.tar.gz with Maven's llvm-firtool JAR package (version 1.137.0) to avoid GLIBC 2.36/2.38 dependency issues on older systems (e.g. Ubuntu 22.04 LTS).

This is based on the latest upstream/main and uses the current firtool version (1.137.0).

Changes:
- Switch from tar.gz static binary to Maven JAR distribution (same version)
- Update build_file_content to extract firtool from JAR structure
- Fixes compatibility with Ubuntu 22.04 (GLIBC 2.35) and similar systems
- Uses firtool-1.137.0 to match upstream version

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -199,11 +199,17 @@ register_toolchains(
 http_archive(
     name = "circt",
     build_file_content = """
-exports_files(glob(["bin/*"]), visibility = ["//visibility:public"])
+package(default_visibility = ["//visibility:public"])
+
+filegroup(
+    name = "bin/firtool",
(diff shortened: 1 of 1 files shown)


oharboe:
@MrAMS Note that the state of CI on this project isn't great: there isn't enough space to run the tests that have been set up. Any contributions to whittle down the tests so CI works here again is generously accepted, though the main problem is that MODULE.bazel pulls in binaries that break the 1GB limit that CI has in this project.

I'm running some local tests on your PR.

PR #425, opened 2025-12-15
  Fix missing preprocessLib.py dependency in makefile_yosys

MrAMS:
## Problem

  Builds were failing with the following error during synthesis:

  /usr/bin/python3: can't open file '.../external/bazel-orfs++orfs_repositories+docker_orfs/OpenROAD-flow-scripts/flow/util/preprocessLib.py': [Errno 2] No such file or directory
  make: *** [external/bazel-orfs++orfs_repositories+docker_orfs/OpenROAD-flow-scripts/flow/Makefile:199: bazel-out/k8-fastbuild/bin/objects/asap7/RegFile_0_5/base/lib/asap7sc7p5t_AO_RVT_FF_nldm_211120.lib] Error 2

  ## Root Cause

  The MAKEFILE_SHARED variable in docker.BUILD.bazel defined shared Makefile dependencies but only included scripts/*.py, missing the utility scripts from the util/ directory. This caused makefile_yosys to lack access to preprocessLib.py and other utility scripts required during the synthesis process.

  ## Solution

  This PR updates MAKEFILE_SHARED to include:
  - OpenROAD-flow-scripts/flow/util/*.py - Python utility scripts including preprocessLib.py, merge_lib.py, etc.
  - OpenROAD-flow-scripts/flow/util/*.sh - Shell utility scripts

  By consolidating all shared dependencies in MAKEFILE_SHARED, both makefile and makefile_yosys now have access to all required utility scripts, eliminating code duplication and ensuring consistency.

  ## Testing

  Verified that the synthesis build completes successfully:
  - Cleaned build cache and rebuilt from scratch
  - Confirmed preprocessLib.py is now accessible in the sandbox
  - Build completed successfully with all synthesis stages passing

### DIFF
--- a/docker.BUILD.bazel
+++ b/docker.BUILD.bazel
@@ -103,6 +103,8 @@ filegroup(
 
 MAKEFILE_SHARED = [
     "OpenROAD-flow-scripts/flow/util/utils.mk",
+    "OpenROAD-flow-scripts/flow/util/*.py",
+    "OpenROAD-flow-scripts/flow/util/*.sh",
     "OpenROAD-flow-scripts/flow/scripts/*.py",
     "OpenROAD-flow-scripts/flow/scripts/*.sh",
     "OpenROAD-flow-scripts/flow/scripts/*.yaml",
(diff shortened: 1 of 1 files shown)


oharboe:
preprocessLib.py has been retired from ORFS for some time, you must be picking this up from some repository that is using an *old* version of ORFS. Run `bazelisk run @bazel-orfs//:bump` in the repository you are using bazel-orfs from....

The change you make here makes sense, but your PR is not a single concern, it includes that other fix you are working on...

MrAMS:
Thank you for telling me more about this issue.
You are right, I forget to exclude other changes because I cannot run it locally without them, I will fix them later.

PR #431, opened 2025-12-16
ci: avoid overprovisioning, faster builds

oharboe:
@MrAMS :crossed_fingers:

PR #432, opened 2025-12-16
ci: add Bazel caching with bazel-contrib/setup-bazel

MrAMS:
- Add bazel-contrib/setup-bazel@0.15.0 to all CI jobs for optimized caching
- Use job-specific cache keys for regular jobs (workflow-job)
- Use matrix-specific cache keys for matrix jobs (workflow-target)
- Enable bazelisk-cache, disk-cache, and repository-cache for all jobs
- Replace manual bazelisk installation with setup-bazel action in preinstalled-orfs job

related [pull#431](https://github.com/The-OpenROAD-Project/bazel-orfs/pull/431)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,13 +17,14 @@ jobs:
     env:
       DEBIAN_FRONTEND: "noninteractive"
     steps:
-      - name: Download buildifier
-        run: |
-          wget https://github.com/bazelbuild/buildtools/releases/download/v6.4.0/buildifier-linux-amd64 -O /usr/local/bin/buildifier
-          chmod +x /usr/local/bin/buildifier
-          buildifier -version
       - name: Checkout bazel-orfs
         uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


oharboe (changes requested):
Neat!

[Computer says no](https://www.youtube.com/watch?v=0n_Ty_72Qds) so far though.

MrAMS:
Oh yeeh, you can check the ci output, it hit the cache :rocket: :rocket: :rocket:

oharboe:
> Sorry for the spam! I was trying to debug the CI, but fixing the local flow is turning out to be harder than expected. I’m going to roll back to the previous version for single `local flow` (without cache) for now.

Yes, so many things to fix... It requires a lot of willpower to leave it all be and fix one thing at the time.

oharboe:
> I see 'local flow' is not the slowest (`smoketests` with cache costs ~ 7min), so this trade off make sense. 🥳

I'm not finished being excited about having CI working at all at this point. Thanks!

Issue #443, opened 2025-12-17
Somewhat flaky reading in of variables.yaml


oharboe:
This happens for mysterious reasons in more recent versions of Bazel.

Is this a problem in bazel-orfs, or can we upgrade something to make this go away?

```
ERROR: error loading package 'sram': at /home/runner/work/bazel-orfs/bazel-orfs/openroad.bzl:12:6: Encountered error while reading extension file 'json.bzl': no such package '@@+_repo_rules2+orfs_variable_metadata//': no such package '@@+orfs_repositories+docker_orfs//': no such package '@@+orfs_repositories+com_github_nixos_patchelf_download//': java.io.IOException: Error downloading [https://github.com/NixOS/patchelf/releases/download/0.18.0/patchelf-0.18.0-x86_64.tar.gz] to /home/runner/.bazel/external/+orfs_repositories+com_github_nixos_patchelf_download/temp5417613084740512892/patchelf-0.18.0-x86_64.tar.gz: GET returned 504 Gateway Time-out
Loading: 0 packages loaded
```

PR #445, opened 2025-12-17
feat: add experimental_repository_downloader_retries

MrAMS:
maybe this will help [The-OpenROAD-Project/bazel-orfs/issues/443](https://github.com/The-OpenROAD-Project/bazel-orfs/issues/443)


oharboe:
Will reopen and merge to try to fix things if the problem persists, running with 8.5.0 for a while first...

PR #449, opened 2025-12-17
rules_python: 1.7.0

oharboe:


MrAMS:
> @MrAMS Do you understand what is wrong here???

I am busy with something else, I will figure it out later.

PR #451, opened 2025-12-17
test: CI test with empty commit

oharboe:
@MrAMS Testing empty commit to see that CI works...


oharboe:
@MrAMS CI seems to work on master...

PR #456, opened 2025-12-18
ci: speed up free-disk-space action

MrAMS:
It can reduce time from ~4min to 47s :rocket: which can still save ~20GB which is enough for us.
I think it is a good trade off.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -170,9 +170,10 @@ jobs:
           android: true
           dotnet: true
           haskell: true
-          large-packages: true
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -170,9 +170,9 @@ jobs:
           android: true
           dotnet: true
           haskell: true
```
oharboe: add comments on tradeoffs here for future maintainers

oharboe (changes requested):
comment to document trade-offs

PR #461, opened 2025-12-18
chisel_test: classical macro snag

oharboe:
The macro is evaluated where it is used, so the macro was working inside bazel-orfs and broken outside.

@MrAMS FYI, classic snag.


oharboe:
@MrAMS This is pretty vile stuff! :cowboy_hat_face:

PR #465, opened 2025-12-23
Optuna minimum clock period

oharboe:
@MrAMS First simple idea on how to allow command line override of config.mk variables. Simple Optuna script next.

### DIFF
--- a/openroad.bzl
+++ b/openroad.bzl
@@ -628,9 +628,14 @@ def _verilog_arguments(files, short = False):
         ),
     }
 
-def _config_content(arguments, paths):
+def _config_content(ctx, arguments, paths):
+    defines_for_stage = {
+        var: value
+        for var, value in ctx.var.items()
+        if var in ALL_STAGE_TO_VARIABLES[ctx.attr._stage]
+    }
     return "".join(
-        sorted(["export {}?={}\n".format(*pair) for pair in arguments.items()]) +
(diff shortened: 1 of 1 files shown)


oharboe:
@MrAMS See latest commit, much faster scan of e.g. place/cts parameters as preceeding stages are not re-build.

oharboe:
Can we make this a generic feature that can be run on any orfs_flow()?

That would be really neat...

MrAMS:
That would be a great goal, but I suggest we approach it in stages. I propose we first merge the simple example in [PR #466](https://github.com/The-OpenROAD-Project/bazel-orfs/pull/466) (or the basic functionality provided in this PR). Once that is in place, we can build additional DSE demos to explore what the optimal design actually looks like.

To be honest, I suspect making this feature fully generic might be difficult. The requirements for Design Space Exploration (DSE) can be very diverse, and the interfaces for passing parameters to hardware generators (like Chisel, which I am [currently investigating](https://github.com/MrAMS/bazel-chisel-verilator-openroad-demo/commit/bbc4a5cfe8d55b7495debc1d3081ded43ea17919)) vary significantly. It might be better for orfs to provide a foundational interface rather than trying to cover every use case out of the box.

oharboe:
> To be honest, I suspect making this feature fully generic might be difficult.

I think you are absolutely right. Optuna doesn't come with "standard" optimization functions, it must be implemented by the user to express intent.

I think, similarly an Python Optuna script + a user bazel flow is this optimization function that must be implemented by the user.

PR #466, opened 2025-12-24
Add optuna DSE example for CORE_UTILIZATION and PLACE_DENSITY

MrAMS:
@oharboe check this demo.


MrAMS:
image

image


image

optuna/README.md
```diff
@@ -1,16 +1,124 @@
-# Optuna minimum example
+# Optuna-Based Design Space Exploration (DSE)
 
-Use `--define X=Y` to override variables in ORFS `config.mk`. `--define X=Y` is adequate for simple ORFS flows to drive Optuna where there's no ambiguity as to which of several subflows that varibles such as `PLACE_DENSITY` refers to:
+Intelligent parameter optimization using Optuna to find optimal design parameter combinations that meet timing constraints.
 
-    bazelisk run --define PLACE_DENSITY=0.1234 //:lb_32x128_floorplan_deps /tmp/x print-PLACE_DENSITY
+## Experiment Design
```
oharboe: add screenshots, looks good on github README.md :-)

optuna/README.md
```diff
@@ -1,16 +1,124 @@
...
 
-Outputs:
+**Goal**: Find optimal CORE_UTILIZATION and PLACE_DENSITY at a fixed clock frequency to minimize area (or area + power).
 
-    PLACE_DENSITY = 0.1234
+**Why Optuna?**
```
oharboe: - There are many Python hyperparameter search frameworks, Optuna shines in simplicity and good results for limited effort. No need install or configure a server for remote execution.

oharboe:
gemini summary. we should do trials in parallel.


Here is a concise, high-level summary formatted in Markdown, ready to copy-paste into a GitHub Pull Request review. It flags the issue without writing the code for them.
🧱 Blocked: Parallel Execution & Bazel Locks
To support running Optuna with n_jobs > 1, we need to address the default Bazel server lock which prevents concurrent builds in the same workspace.
Required Changes:
 * [ ] Isolate Trials: Please update the command construction to specify a unique --output_base for each trial (e.g., derived from trial.number).
 * [ ] Flag Placement: Ensure --output_base is passed as a startup option (before the build command), while --define args remain after.
 * [ ] Resource Management: Since this will generate a fresh cache for every trial, please add a cleanup step to remove the temporary output directory after the metric is collected.
Why this works for a review:
 * Clear Heading: Uses an emoji and clear title to indicate this is a blocking architectural issue, not a nitpick.
 * Checklist: Uses - [ ] syntax so the reviewee can tick off items as they fix them.
 * Distinction of Flags: It explicitly points out the location of the flag (Startup option vs Command option) which is the most common error in this specific implementation, without writing the python code for them.
Would you like me to draft a comment on how to handle the disk-space cleanup specifically?

oharboe:
Here is the updated summary for the GitHub review, adding the requirement for file-based persistence (SQLite) to ensure failures and progress are saved.
💾 Persistence & Fault Tolerance
To ensure we don't lose the study history if the process crashes—and to properly track the failed trials we expect from the Bazel integration—we need to move away from the default in-memory storage.
Required Changes:
 * [ ] Add SQLite Backend: Please switch optuna.create_study to use a file-based storage URL (e.g., storage="sqlite:///bazel_opt.db").
 * [ ] Resume Capability: Enable load_if_exists=True so we can resume the optimization or inspect the history of a study that was interrupted.
 * [ ] Failure Tracking: Ensure the storage configuration is set up so that TrialPruned or execution errors are committed to the DB immediately, allowing us to debug specific failed Bazel commands later.
Would you like me to create a "suggested change" code block for the create_study line to include in the review?

optuna/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+Outputs:
+
+    PLACE_DENSITY = 0.1234
+
+## Simple Optuna example to search for minimum clock period
+
+Search for minimum clock period using the synthesis stage only:
+
+    bazelisk run //optuna:min-clock-synth
```
oharboe: Doesn't work:

```
ERROR: Skipping '//optuna:min-clock-synth': no such target '//optuna:min-clock-synth': target 'min-clock-synth' not declared in package 'optuna' defined by /home/oyvind/bazel-orfs/optuna/BUILD
ERROR: no such target '//optuna:min-clock-synth': target 'min-clock-synth' not declared in package 'optuna' defined by /home/oyvind/bazel-orfs/optuna/BUILD
```

PR #472, opened 2025-12-25
optuna: avoid python language servers from locking up with 100% cpu when discovering bazel folders

oharboe:
@MrAMS FYI


MrAMS:
Thanks! :+1:

MrAMS:
I found that there is no perfect solution for `rules_python` to work well with IDE. Now I use the fake venv instead. :cry:

PR #475, opened 2025-12-25
[DRAFT] optuna: switch to mock-cpu

oharboe:
A mock multicore CPU in SystemVerilog to keep the Chisel concern out of the bazel-orfs scaling for Optuna issues at first.

@MrAMS Thoughts?

Needs `VERILOG_TOP_PARAMS` + some study work.

My thinking is that bazel-orfs needs to demonstrate that it can solve the compute scaling of these studies and that this will require some collaboration and fixes in bazel-orfs itself...

### DIFF
--- a/optuna/BUILD
+++ b/optuna/BUILD
@@ -10,14 +10,12 @@ exports_files([
 
 # DSE-specific flow configuration with tighter constraints
 orfs_flow(
-    name = "lb_32x128",
+    name = "mock-cpu",
     abstract_stage = "cts",
     arguments = {
-        # Tight timing constraints for meaningful DSE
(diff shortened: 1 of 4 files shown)


MrAMS:
I fully agree with your perspective. It makes sense to treat flow tuning and architectural exploration as separate stages. We can restructure the optuna directory to have two separate subfolders: one dedicated to sweeping internal EDA parameters (like `CORE_UTILIZATION` and `PLACE_DENSITY`) to establish a baseline, and another for the actual DSE of RTL architectural parameters (systemverilog for now, we can add thrid subfolder for chisel in the future).

PR #476, opened 2025-12-27
Dse params

oharboe:
@MrAMS Step 1 implemented and tested over a coffee... Thoughts?

Will tinker with it more later, but have few cycles in next days.

### PR Description: Implement Multi-Config Transitions for Parallel DSE

This PR introduces the foundational Starlark transitions required to break away from global `--define` flags. By moving to a transition-based configuration, we can now "fork" the build graph into  parallel flows within a single invocation, solving the ambiguity problem for concurrent DSE trials.

#### Changes:

* **DSE Build Settings:** Introduced custom Starlark build settings to hold trial-specific configuration strings.
* **1-to-N Transition Logic:** Added a transition that allows a parent rule to spawn multiple `orfs_flow` instances, each with its own unique configuration context.
* **Provider Updates:** Updated the flow providers to propagate these configuration hashes, ensuring Bazel treats each trial as a unique node in the action graph.

#### The Roadmap:

1. **Stage-Specific Filtering (Next Step):** I’ll be updating `orfs_flow()` to ingest a `.json` parameter blob. A specialized action will "whittle down" this file into stage-specific fragments. This ensures that a change to `STRATEGY` in the routing stage doesn't invalidate a 4-hour placement result.
2. **Trial Mapping:** Implement the `orfs_flow(name="flow_{n}")` naming convention where trial  maps to slot .
3. **Optuna Orchestration:** The final piece is a Python glue layer using `concurrent.futures`. Optuna will manage the Bayesian state, wait for a batch of  trials to fill (or the final trial of a study), and then assemble a single `bazelisk build //flows:all` command.

This effectively turns Bazel into our high-performance execution engine for Bayesian search, keeping the "heavy lifting" in the build farm while Optuna handles the "intelligence" in the orchestrator.

### DIFF
--- a/dse/BUILD
+++ b/dse/BUILD
@@ -1,0 +1,20 @@
+load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
+load("//:param.bzl", "param")
+
+string_flag(
+    name = "density",
+    build_setting_default = "50",
+)
+
+string_flag(
+    name = "utilization",
+    build_setting_default = "45",
+)
+
+param(
+    name = "params",
+    parameters = {
+        ":density": "PLACE_DENSITY",
+        ":utilization": "CORE_UTILIZATION",
+    },
+)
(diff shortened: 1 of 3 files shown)


oharboe:
I think this can be done more straightforwardly

PR #478, opened 2025-12-28
Dse build settings

oharboe:
@MrAMS Check this out... `--define X=Y` approach makes it impossible to disambiguate between multiple orfs_flow() in graphs, with BuildSettingsInfo you can.

So the next thing to try is to create N orfs_flow() and run them in parallel with Optuna and framework invert Optuna studies.

How hard can it be? :smile_cat:

### DIFF
--- a/dse/README.md
+++ b/dse/README.md
@@ -1,0 +1,70 @@
+# Design Space Exploration (DSE) Parameterization with Bazel
+
+This project demonstrates how to use **Bazel build settings** with orfs_flow() `settings` argument to manage design parameters for hardware design space exploration (DSE).
+
+## Use Case
+
+Suppose you want to sweep or optimize parameters such as **core utilization** and **placement density** in your hardware design flow. This setup allows you to:
+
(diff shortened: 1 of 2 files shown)


oharboe:
Yes, bur use list comprehensions in BUILD files to generate many orfs_flow()s.

The param.bzl was a dead end, removed.

dse/README.md
```diff
@@ -0,0 +1,70 @@
...
+   )
+   ```
+
+2. **Convert Parameters to JSON for Flow Graphs**
+
+   The system converts configuration parameters to a `.json` file that can be fed into a static `orfs_flow()` graph. Each stage in the flow runs an action to whittle the `.json` file down to just the parameters needed for that stage.
+
+   This allows configuring the entire `orfs_flow()` graph directly from the Bazel command line using `--//path:parameter=value` pairs defined by the user.
+
+3. **Map Build Settings to Parameter Names**
```
MrAMS: so should be removed?

PR #481, opened 2025-12-31
optuna: prune to simplest interesting case

oharboe:
multiobjective driven by synthesis parameters is the minimum interesting case, prune everything else and reduce size of python code and dependencies in terms of what a reader has to understand.

Fancy dependencies to visualize, if they are easy to use and have low cognitive load, such as generating interactive 3d interactive HTML for pareto fron is fine.

The goal is the smallest possible interesting example that demonstrates and tests key features in bazel-orfs so that the user can take this as a starting point to create his own DSE.


oharboe:
@MrAMS FYI

PR #491, opened 2026-01-07
feat: remove --jobs 1 restriction and use Bazel resource management

MrAMS:
Instead of forcing --jobs 1 for non-synth stages, remove this restriction entirely and let Bazel manage parallelism based on available resources.

Users can now control resource usage via Bazel flags:
  - --local_ram_resources: Limit based on available memory
  - --jobs: Set max parallel jobs globally
  - --local_cpu_resources: Control CPU allocation

This allows much better CPU utilization during DSE while still avoiding OOM through Bazel's intelligent resource scheduling.

Benefits:
- Better CPU utilization across all build stages
- Avoids manual per-stage job limits
- Flexible resource management via standard Bazel flags
- Significantly faster DSE with parallel builds

### DIFF
--- a/openroad.bzl
+++ b/openroad.bzl
@@ -165,6 +165,19 @@ def orfs_environment(ctx):
         "WORK_HOME": _work_home(ctx),
     }
 
+def _thread_env(ctx):
+    """Allow Bazel defines to override ORFS NUM_CORES.
+
+    Checks in priority order:
+    1. NUM_CORES (standard ORFS variable)
+    2. ORFS_NUM_CORES (alternative name)
+    3. OPENROAD_THREADS (legacy name)
+    """
+    threads = ctx.var.get("NUM_CORES") or ctx.var.get("ORFS_NUM_CORES") or ctx.var.get("OPENROAD_THREADS")
+    if threads:
+        return {"NUM_CORES": str(threads)}
+    return {}
+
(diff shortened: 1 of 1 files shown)


oharboe (commented):
I'd love to find a better solution...

openroad.bzl
```diff
@@ -165,6 +165,13 @@ def orfs_environment(ctx):
         "WORK_HOME": _work_home(ctx),
     }
 
+def _thread_env(ctx):
+    """Allow Bazel defines/action envs to override ORFS NUM_CORES."""
+    threads = ctx.var.get("ORFS_NUM_CORES") or ctx.var.get("OPENROAD_THREADS")
```
oharboe: Why is this needed?

Doesn't `--define NUM_CORES=N` work today?

MrAMS: Yes

oharboe: ref github issue, I think we should fix the bug, it isn't working as intended.

oharboe:
> I added `NUM_CORES`.

--define should work for all variables in variables.yaml, which includes NUM_CORES. I prefer fixing the bug to make the system work as intended so we don't get the complexity of workarounds.

oharboe:
Close?

--define NUM_CORES=42 now works on main

PR #493, opened 2026-01-08
Config override

oharboe:
@MrAMS Doesn't work yet, I missed a spot or something... But a fix along these lines is what I have in mind.


oharboe:
merged

PR #495, opened 2026-01-08
orfs_flow: config override

oharboe:
bazelisk run //:tag_array_64x184_synth_deps --define NUM_CORES=42 /tmp/x print-NUM_CORES

Should print out NUM_CORES=42

Fixed.

@MrAMS FYI


MrAMS:
Great! Thank you :partying_face:

PR #503, opened 2026-02-11
add worker wrapper to fir_library

PeterAaser:
Intenteded to provide a way to limit the number of running job
instances of a certain type by using Bazel's worker mechanism.

https://bazel.build/remote/persistent#number-of-workers

This is less than ideal, hopefully more fine grained control
over job scheduling will be provided in the future:

https://github.com/bazelbuild/bazel/issues/27950
https://github.com/bazelbuild/bazel/pull/28013

In short, this mechanism allows us to add
--worker_max_instances=FirGeneration=N to a bazel invocation.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -565,3 +565,9 @@ py_binary(
     srcs = ["bsp.py"],
     visibility = ["//visibility:public"],
 )
+
+py_binary(
+    name = "worker_wrapper",
+    srcs = ["worker_wrapper.py"],
+    visibility = ["//visibility:public"],
+)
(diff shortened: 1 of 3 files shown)


oharboe (changes requested):
Hmmm.... this seems to fix one problem and introduce new more complex problems that will hit us in production...

1. what is the situation today?
2. what problem are we solving?
3. What is being proposed here?
4. What are the pros and cons of this solution?


The worker_wrapper.py is fraught with imponderables w.r.t. startup/shutdown, ctrl-c, combining stdout/stderr, etc.

PR #532, opened 2026-03-12
build: add resource_set to prevent CPU oversubscription

oharboe:
Parallel bazel builds cause 370x slowdowns for OpenROAD actions because bazel defaults to 1 CPU per action while OpenROAD spawns threads for all cores. With 16 concurrent actions × 16 threads = 256 threads on 16 cores, context switching and cache thrashing dominate.

Evidence: tag_array_64x184 global placement (379 cells) takes 1.4s alone but 521s inside `bazelisk test ...`. Same computation, same output sha.

Fix: declare resource_set on every ctx.actions.run_shell() call. Multi-threaded stages (place, cts, grt, route) request all CPUs so bazel runs at most one at a time. Single-threaded stages (synth, floorplan, final) request 1 CPU, allowing full parallelism.

A repository_rule (detect_cpus) runs `nproc` once at workspace setup to determine the host CPU count. This is invisible to users — no .bazelrc flags, no env vars, no per-design configuration needed.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -56,6 +56,13 @@ python.toolchain(
 )
 use_repo(python, "python_3_13_host")
 
+detect_cpus = use_repo_rule("//:detect_cpus.bzl", "detect_cpus")
+
+detect_cpus(
+    name = "host_cpus",
+    python_interpreter = "@python_3_13_host//:python",
+)
+
 pip = use_extension("@rules_python//python/extensions:pip.bzl", "pip")
 pip.parse(
     hub_name = "bazel-orfs-pip",
--- a/detect_cpus.bzl
+++ b/detect_cpus.bzl
@@ -1,0 +1,32 @@
+"""Repository rule to detect the number of CPUs available on the host.
+
+This runs once during workspace setup and writes a .bzl file with the
+CPU count, which openroad.bzl uses for resource_set scheduling.
+"""
+
+def _detect_cpus_impl(repository_ctx):
+    python = repository_ctx.path(repository_ctx.attr.python_interpreter)
+    result = repository_ctx.execute([
+        str(python),
+        "-c",
+        "import os; print(os.cpu_count() or 1)",
+    ])
+    if result.return_code == 0:
+        cpus = int(result.stdout.strip())
+    else:
+        cpus = 8
+    repository_ctx.file("BUILD.bazel", "")
+    repository_ctx.file("cpus.bzl", "NUM_CPUS = %d\n" % cpus)
+
+detect_cpus = repository_rule(
+    implementation = _detect_cpus_impl,
+    attrs = {
+        "python_interpreter": attr.label(
+            mandatory = True,
+            allow_single_file = True,
+            doc = "Hermetic Python interpreter used to detect CPU count.",
+        ),
+    },
+    local = True,
+    configure = True,
+)
--- a/docs/resource_scheduling.md
+++ b/docs/resource_scheduling.md
@@ -1,0 +1,158 @@
+# Resource Scheduling for Bazel Actions
+
+## The problem
+
+Bazel defaults to 1 CPU per action, allowing many actions to run in parallel.
+OpenROAD stages that spawn internal threads suffer severe slowdowns from this:
+context switching and cache thrashing dominate when N actions each spawn
+NUM_CORES threads on NUM_CORES physical cores.
+
+Evidence: `tag_array_64x184` global placement (379 cells) takes 1.4s alone but
+521s inside `bazelisk test ...`. Same computation, same output — 370x slower.
+
+## The fix
+
+Every `ctx.actions.run_shell()` call declares a `resource_set`. Multi-threaded
+stages request all CPUs so Bazel runs at most one at a time. Single-threaded
+stages request 1 CPU, allowing full parallelism.
+
+A `repository_rule` (`detect_cpus`) runs hermetic Python's `os.cpu_count()`
+once at workspace setup to determine the host CPU count. This is written to
+`@host_cpus//:cpus.bzl` as `NUM_CPUS` and loaded by `openroad.bzl`.
+
+## Stage threading classification
+
+OpenROAD is always invoked with `-threads NUM_CORES`, but ORFS's `load.tcl`
+immediately sets `sta::set_thread_count 1` (workaround for flaky STA in
+rel_3.0+). Only algorithms with their own internal threading (OpenMP, etc.)
+actually use multiple cores.
+
+### Multi-threaded stages (request all CPUs)
+
+| Stage   | Make target    | Multi-threaded sub-stages                                                                           |
+|---------|----------------|------------------------------------------------------------------------------------------------------|
+| place   | `do-place`     | `global_placement` (RePlAce, OpenMP), `global_placement -incremental` (resize), `detailed_placement` |
+| cts     | `do-cts`       | `clock_tree_synthesis` (TritonCTS), `detailed_placement`, `repair_timing_helper`                      |
+| grt     | `do-5_1_grt`   | `global_route` (FastRoute), `repair_design_helper`, `repair_timing_helper`, `detailed_placement`     |
+| route   | `do-5_2_route` | `detailed_route` (TritonRoute, heavily threaded)                                                     |
+
+### Single-threaded stages (request 1 CPU)
+
+| Stage             | Make target            | Sub-stages                                              |
+|-------------------|------------------------|---------------------------------------------------------|
+| synth             | `do-yosys`             | Yosys synthesis (single-threaded)                       |
+| floorplan         | `do-floorplan`         | `floorplan`, `macro_place`, `tapcell`, `pdn`            |
+| final             | `do-final`             | `density_fill`, `final_report`, KLayout GDS merge       |
+| generate_abstract | `do-generate_abstract` | `write_timing_model`, `write_abstract_lef`              |
+
+### Empirical thread counts (tag_array_64x184, ASAP7, 16-core host)
+
+Measured via `ps -eo nlwp` sampling at 200ms intervals:
+
+| TCL script              | Max threads observed | Samples |
+|-------------------------|---------------------:|--------:|
+| `global_place_skip_io`  |                   16 |     146 |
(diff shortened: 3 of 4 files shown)


oharboe:
@maliberty FYI conclusion after tinkering with how to handle throttling in bazel and OpenROAD.

# Proposal: Self-Throttling OpenROAD Binary

## Problem

When Bazel runs multiple OpenROAD actions in parallel, each instance spawns
`NUM_CORES` threads regardless of how many siblings are running. With 16
concurrent actions on a 16-core machine, this creates 256 threads competing for
16 cores — a 370x slowdown from context switching and cache thrashing.

## Tested and discarded: Bazel `resource_set` approach

A `resource_set` solution was implemented and tested in bazel-orfs. It works by
declaring CPU and memory needs per Bazel action so the scheduler avoids
oversubscription:

- A `repository_rule` detects CPU count at workspace setup via hermetic Python
- Each stage rule declares `resource_set` — multi-threaded stages (place, cts,
  grt, route) request all CPUs; single-threaded stages request 1 CPU
- Bazel's scheduler then runs at most one heavy stage at a time

**This fixed the 370x oversubscription problem**, but created new ones:

- **Coarse granularity**: each stage (e.g., `do-place`) mixes multi-threaded
  substages (`global_placement`, 16 threads) with single-threaded ones
  (`io_placement`, 1 thread). The entire stage holds all CPUs even during
  the single-threaded parts, blocking other work.
- **Breaking into substages is impractical**: splitting stages into separate
  Bazel actions for finer resource control causes artifact explosion (each
  substage produces a hundreds-of-MB ODB file) and creates white-box coupling
  with ORFS internals (the `do-` substage targets are explicitly "subject to
  change" per the ORFS Makefile).
- **Bazel-specific**: the fix only helps when building via Bazel. Direct `make`
  invocations, CI scripts, and other build systems get no benefit.
- **Requires external metadata**: bazel-orfs must know which stages are
  multi-threaded — knowledge that belongs in OpenROAD/ORFS, not the build
  system.

This motivated exploring a solution inside OpenROAD itself.

## Proposed solution: throttling sections

OpenROAD gains awareness of sibling instances and self-throttles at section
boundaries.

### API sketch (TCL)

```tcl
# Before a heavy algorithm:
throttle::begin -cpu 16 -memory 4096   ;# "I'd like 16 cores and ~4GB"

global_placement ...                    ;# runs with negotiated resources

throttle::end                           ;# back to baseline resources
```

### Default behavior

When no throttling section is active, OpenROAD presents as a lightweight
process — approximately 1 CPU and memory comparable to a single C++ compiler
invocation. This makes it an **idiomatic Bazel citizen**: Bazel's default
assumption of 1 CPU per action is correct for most of OpenROAD's runtime
(loading designs, writing files, reporting). Only the heavy algorithms
(placement, routing) need more.

### Sibling detection and negotiation

When entering a throttling section, OpenROAD:

1. **Discovers siblings** via a shared coordination mechanism (see options below)
2. **Declares resource needs** — CPU count and approximate memory
3. **Receives allocation** — a fraction of available resources based on how many
   siblings are also in heavy sections
4. **Configures internal threading** — sets `omp_set_num_threads()` and internal
   thread pool sizes to the allocated count
5. **Releases resources** on `throttle::end` — siblings waiting for resources
   can now scale up

### Coordination mechanism: POSIX shared memory + semaphore

The recommended coordination backend is POSIX shared memory (`shm_open`) with
a named semaphore (`sem_open`). A shared memory segment holds a slot table:

```c
struct throttle_state {
    int num_cpus;              // detected at startup
    int num_slots;             // MAX_CONCURRENT_HEAVY
    struct {
        pid_t pid;
        int cpu_requested;
        int memory_mb;
        enum { IDLE, HEAVY } state;
    } slots[MAX_SLOTS];
    sem_t mutex;               // protects the table
};
```

Why POSIX shared memory:

- **No files, no conventions** — lives in `/dev/shm`, kernel-managed, no temp
  directories to agree on, no stale files to clean up
- **Atomic** — `sem_wait`/`sem_post` on a named semaphore, no race windows
- **Graceful fallback** — if `shm_open` fails (sandboxed environment, no
  `/dev/shm`, macOS restrictions), log "throttling not available" and fall back
  to `-threads N` behavior. Pure no-op, zero breakage.
- **Low cognitive load** — no environment variables to set, no jobserver pipes
  to configure, no coordination directories to manage. Multiple OpenROAD
  processes on the same host just work.

### Extensibility

The `throttle::begin` / `throttle::end` API is deliberately backend-agnostic.
The initial implementation uses POSIX shared memory, but the same interface
can support alternative backends as use-cases justify them:

- **GNU Make jobserver** (`--jobserver-auth` in `MAKEFLAGS`) — for integration
  with Make/Ninja/Cargo ecosystems where the parent process manages the token pool
- **Shared file/lockfile protocol** — for environments without `/dev/shm`
- **cgroup-aware throttling** — read container CPU/memory limits and adjust

These can be added incrementally behind command-line options (e.g.,
`-throttle-backend jobserver`) without changing the TCL-level API or the
section annotations in ORFS scripts.

### Memory-aware throttling

Beyond CPU throttling, OpenROAD knows its own memory usage via `/proc/self/statm`
or `getrusage()`. If memory pressure is detected (e.g., approaching cgroup
limits or a configurable threshold), it can:

1. Reduce thread count to lower per-thread memory overhead
2. Defer memory-intensive operations
3. Signal siblings to also reduce their footprint

This is particularly important for `grt` (global routing) which is both
CPU-intensive and memory-hungry on large designs.

### Relationship to Bazel resource_set

With self-throttling, bazel-orfs no longer needs per-stage `resource_set`
declarations. All OpenROAD actions can declare `{"cpu": 1, "memory": 250}` (the
default), and OpenROAD internally scales up when safe. This:

- Eliminates the need for `detect_cpus` repository rule
- Removes the `_MULTI_THREADED_STAGES` classification
- Makes OpenROAD work correctly with any build system, not just Bazel
- Allows fine-grained per-substage throttling without splitting Bazel actions

### CLI interface

```
openroad -threads auto        # default: self-throttle (new behavior)
openroad -threads 16          # legacy: fixed thread count, no throttling
openroad -threads auto -throttle-dir /tmp/my-build  # custom coordination dir
```

The `-threads auto` mode would be the new default. Existing `-threads N`
behavior is preserved for backward compatibility.

## Precedents

| System | Mechanism | Scope |
|--------|-----------|-------|
| GNU Make jobserver | Token pipe shared with children | Parent-orchestrated |
| OpenMP `OMP_DYNAMIC` | Runtime thread adjustment per region | Single process |
| PostgreSQL `work_mem` | Per-query memory budget, planner throttles | Server-wide |
| Intel TBB task arenas | Global thread pool, dynamic migration | Single process |
| Linux cgroups | Kernel-enforced CPU/memory limits | Container/system |
| Cargo/Rustc | Jobserver client for parallel codegen | Build system |

The proposed OpenROAD approach is unique in combining **peer-to-peer sibling
detection** with **per-section resource declaration** — no parent orchestration
required.

## Status: prototyped, creates new problems

This approach was prototyped and the core mechanism works — OpenROAD
successfully detects siblings and throttles thread counts. However, it fixes
the oversubscription problem while creating others:

- **Non-determinism**: thread counts vary between runs depending on what
  siblings happen to be active at section entry. This makes builds
  non-reproducible and complicates debugging (different timing, different
  memory peaks, potentially different optimization outcomes in
  timing-driven algorithms).
- **Coordination overhead**: scanning `/proc` or a shared directory on every
  `throttle::begin` adds latency. For short substages this overhead dominates.
- **Stale state and races**: siblings can start or exit between detection and
  thread configuration. A process that detects 4 siblings and takes 4 cores
  may find itself alone moments later, under-utilizing the machine. The
  reverse (over-committing) is also possible.
- **Complexity in OpenROAD core**: threading decisions are currently simple
  (`-threads N`). Adding runtime negotiation touches OpenMP, TritonRoute's
  thread pool, FastRoute's parallelism, and STA's thread count — each with
  different threading APIs. Getting this wrong causes subtle bugs.
- **Build system mismatch**: Bazel's scheduler has already committed resources
  to the action based on `resource_set`. If OpenROAD internally throttles to
  fewer cores than declared, those cores sit idle. If it scales beyond what
  Bazel allocated, oversubscription returns. The two scheduling systems fight.
- **Memory estimation is unreliable**: declaring memory needs upfront requires
  knowing the design size before processing it. The same algorithm on
  different designs uses vastly different memory. Reactive throttling (check
  usage mid-run) adds more non-determinism.

## Conclusion

The Bazel `resource_set` approach (declaring resource needs per stage at the
build system level) remains the pragmatic solution. It's deterministic, simple,
and lets Bazel's scheduler make globally optimal decisions. Self-throttling in
the binary is appealing in theory but fights the build system rather than
cooperating with it.

The right long-term path is for ORFS to export resource metadata (following the
`variables.yaml` pattern) so build systems can make informed scheduling
decisions — keeping the intelligence in the scheduler, not the binary.

## Implementation phases

### Phase 1: Section annotations (no-op, zero risk)

Add `throttle::begin` / `throttle::end` TCL commands to OpenROAD that:
- Log "Entering heavy section (cpu=16, memory=~4096MB)" /
  "Leaving heavy section" — useful documentation even without throttling
- Accept `-cpu` and `-memory` arguments for future use
- Do **nothing else** — no shared memory, no thread changes
- Wire up calls in ORFS scripts around heavy algorithms

This establishes the API and section boundaries with zero behavior change.
OpenROAD developers annotate their code incrementally.

### Phase 2: POSIX shared memory negotiation

- `shm_open` a shared segment on first `throttle::begin`
- Register in slot table, count siblings in heavy state
- Set `omp_set_num_threads()` to `num_cpus / active_heavy_count`
- Release slot on `throttle::end`
- Fall back to no-op if `shm_open` fails
- Clean up via `atexit` handler

### Phase 3: Memory-aware throttling

- Monitor `/proc/self/statm` during heavy sections
- Reduce thread count if memory usage exceeds threshold or cgroup limit
- Configurable via `-throttle-memory-limit`

### Phase 4: ORFS integration

- Annotate all heavy algorithm calls with `throttle::begin` / `throttle::end`
- Resource profiles per section become structured metadata
- This metadata could also feed into a `stages.yaml` for build system
  consumption (following the `variables.yaml` pattern)

PR #655, opened 2026-04-07
fix: update verilator, rules_verilator, and add rules_verilog

oharboe:
Upgrade to hermetic-compatible versions and use the new rules_verilog package for verilog providers (VerilogInfo), which was extracted from rules_verilator into its own module.

- verilator 5.036.bcr.3 → 5.046.bcr.2
- rules_verilator 0.1.0 → 0.3.1
- rules_verilog 1.1.1 (new dependency)

The older verilator had non-hermetic Python scripts in its build; 5.046.bcr.2 fixes this. The newer rules_verilator depends on rules_verilog for verilog providers, so verilog.bzl now loads VerilogInfo from @rules_verilog instead of the removed @rules_verilator//verilog:providers.bzl.

Fixes https://github.com/The-OpenROAD-Project/bazel-orfs/issues/654


oharboe:
@hzeller Died after 2-3 hours. :disappointed:

oharboe:
Given that CI is so slow, we can no longer afford the O(n^2) CI luxury. We have to fix post-merge. The tests are generally sorted in order of number of affected users.

hzeller:
so, that means this also was not able to (re-)populate the cache ?

oharboe:
you tell me... https://github.com/The-OpenROAD-Project/bazel-orfs/actions/runs/24089762801

PR #656, opened 2026-04-07
pain-relief: add toolchain ideas for yosys and OpenROAD

oharboe:
Document stop-gap Bazel toolchains for yosys and OpenROAD with pre-built binaries hosted on bazel-orfs GitHub Releases. These are pain-relief until both tools appear on the Bazel Central Registry, at which point we retire them.

Covers toolchain types, providers, pre-built hosting, user-applied patches (the push-past-upstream-issues workflow), and migration path.

### DIFF
--- a/ideas/toolchains.md
+++ b/ideas/toolchains.md
@@ -1,0 +1,996 @@
+# Pain-Relief Toolchains and Packages
+
+## Context
+
+Two things make bazel-orfs painful for new users and CI:
+
+1. **Build time**: yosys takes 10-20 minutes from source, OpenROAD
+   takes 30-60+ minutes. Both require a C++20 compiler, cmake, and
+   platform-specific development headers.
+
+2. **Clone size**: ORFS is a 1.1 GB git checkout. bazel-orfs only
+   needs ~830 KB of scripts/makefiles plus PDK platform files, but
+   `git_override` clones the entire repo including 265 MB of unused
+   design examples.
+
+The proper long-term fix is for yosys and OpenROAD to publish
+first-class Bazel modules on the Bazel Central Registry (BCR), with
+pre-built binaries and proper toolchain definitions. That is not
+something we control, and it will take time.
+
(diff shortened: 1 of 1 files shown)


ideas/toolchains.md
```diff
@@ -0,0 +1,996 @@
...
+
+2. **Clone size**: ORFS is a 1.1 GB git checkout. bazel-orfs only
+   needs ~830 KB of scripts/makefiles plus PDK platform files, but
+   `git_override` clones the entire repo including 265 MB of unused
+   design examples.
+
+The proper long-term fix is for yosys and OpenROAD to publish
+first-class Bazel modules on the Bazel Central Registry (BCR), with
+pre-built binaries and proper toolchain definitions. That is not
+something we control, and it will take time.
```
hzeller, line 19: BCR never provides pre-built binaries, only build descriptions.

(Yosys is already there https://registry.bazel.build/modules/yosys )

oharboe: no slang, no local patches

oharboe: as a yosys toolchain?

oharboe: so binaries must still be hosted ...

hzeller: yeah, but what you can do is to have a github release with a static binary in the release. So that makes hosting simple.

oharboe:
@maliberty Thoughts?

If you give me access, claude will do the rest.

hzeller:
before you go with the big hammer solution, I'd try to set up the cache correctly, so that rebuilding really only is happening whenever there are changes.

oharboe:
> before you go with the big hammer solution, I'd try to set up the cache correctly, so that rebuilding really only is happening whenever there are changes.

I think the problem is that there's no automated test to tell me when I'm not "holding it right".

Can we have an automated test of caching being done right?

oharboe:
@hzeller Also, caching is not the only benefit. The C++ options hell bleed into the projects using bazel-orfs. Binaries give isolation.

PR #657, opened 2026-04-07
feat: validate orfs_flow arguments against ORFS variables.yaml

oharboe:
Catch misspelled or unknown variable names at build time instead of letting them silently propagate to runtime. Documents the patching workflow for adding missing variables.


oharboe:
revisit

oharboe:
@MrAMS Nice idea. Good time saver, I think. But needs to be nursed through.

PR #658, opened 2026-04-07
feat: add OrfsInfo.arguments and orfs_arguments rule for computed flow args

oharboe:
Add a mechanism for flow arguments to accumulate as .json files through OrfsInfo.arguments (depset) and flow from stage to stage. Each stage merges inherited .json files into a .mk config at build time, keeping the .mk format as an internal implementation detail.

The new orfs_arguments rule runs a Tcl script to compute arguments (e.g., CORE_UTILIZATION, PLACE_DENSITY) from stage outputs, storing the result as .json in OrfsInfo.arguments. orfs_flow gains an extra_arguments parameter for per-stage .json injection.

Includes a test variant (lb_32x128 auto_util) demonstrating the pattern: synth → orfs_arguments(auto_utilization.tcl) → floorplan.

### DIFF
--- a/docs/orfs_arguments.md
+++ b/docs/orfs_arguments.md
@@ -1,0 +1,137 @@
+# `orfs_arguments` — Computed Flow Arguments
+
+## Overview
+
+`orfs_arguments` is a Bazel rule that runs a Tcl script to compute ORFS
+flow arguments (e.g., `CORE_UTILIZATION`, `PLACE_DENSITY`) from stage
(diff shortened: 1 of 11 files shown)


oharboe:
@MrAMS Thoughts?

oharboe:
@hzeller Caching is still broken somehow...

oharboe:
@MrAMS I think this is a good idea, but I'm letting it lie for now. To be returned to when the need arises...

PR #659, opened 2026-04-07
yosys: build libffi from source for hermetic plugin support

oharboe:
Yosys with ENABLE_PLUGINS=1 requires ffi.h and libffi for DPI call support. Previously this depended on the host having libffi-dev installed, which broke on CI environments without it (fatal error: ffi.h: No such file or directory).

Download and build libffi from source, following the same pattern used for TCL and flex. The header path is added to CXXFLAGS and the library path is exported via LIBRARY_PATH so the linker finds -lffi from yosys's Makefile fallback (used when pkg-config is unavailable).

### DIFF
--- a/gallery/MODULE.bazel
+++ b/gallery/MODULE.bazel
@@ -21,7 +21,6 @@ local_path_override(
     path = "../mock/klayout",
 )
 
-
 # --- Yosys built from source (shared with bazel-orfs) ---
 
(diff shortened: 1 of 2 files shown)


oharboe:
@hzeller FYI

hzeller:
was https://registry.bazel.build/modules/libffi not an option ?

PR #660, opened 2026-04-07
Bazel bcr ify

oharboe:


oharboe:
@hzeller Looks better, but still some libffy shenanigans, by the looks of it.

hzeller:
I still don't know why you attempt to build yosys manually, this looks somewhat fragile.
I'd try to start with yosys and sv-lang on BCR, then apply patches if needed.

PR #661, opened 2026-04-07
Bazel bcr ify 2

oharboe:


oharboe:
@hzeller Better?

chisel/MODULE.bazel
```diff
@@ -25,10 +25,10 @@ local_path_override(
     path = "../mock/klayout",
 )
 
-bazel_dep(name = "bazel-orfs-yosys", version = "0.0.1")
-local_path_override(
-    module_name = "bazel-orfs-yosys",
-    path = "../yosys",
+single_version_override(
```
hzeller, line 30: what are you overriding here ? the bazel_dep() fetching the base yosys is not in this module fuile.

extension.bzl
```diff
@@ -82,19 +71,12 @@ _default_tag = tag_class(
         "yosys_abc": attr.label(
             mandatory = False,
             cfg = "exec",
-            default = Label("@yosys//:yosys-abc"),
+            default = Label("@yosys//:yosys_abc"),
```
hzeller: I'd probably use the BCR abc here (which is the yosys abc) https://registry.bazel.build/modules/abc

patches/yosys-visibility.patch
```diff
@@ -0,0 +1,11 @@
+--- a/BUILD.bazel
++++ b/BUILD.bazel
+@@ -8,7 +8,7 @@
+ load("@rules_python//python:py_binary.bzl", "py_binary")
+ load(":yosys_utils.bzl", "share_dst", "share_tree")
+ 
+-package(default_visibility = ["//visibility:private"])
++package(default_visibility = ["//visibility:public"])
```
hzeller: what needs to be public that isn't already ?

hzeller (commented):
Looks like a good start.
Some questions below.

oharboe:
@hzeller Feedback greatly appreciated... very much a chaotic work in progress. Claude is coding here and your comments provide the required constraints to do this idiomatically. I offer what I can knowing roughly what is needed and knowing the bazel-orfs use-case.

patches/yosys-visibility.patch
```diff
@@ -0,0 +1,9 @@
+--- a/BUILD.bazel
++++ b/BUILD.bazel
+@@ -1,5 +1,6 @@
+ share_tree(
+     name = "yosys_share",
```
hzeller, line 7: what is `yosys_share` used for ?

oharboe: I think some comments were added...

hzeller: there is no comment describing what yosys_share is used for.

oharboe: fixed

oharboe: > what is `yosys_share` used for ?

Various flow files, detailed in comment now.

hzeller:
is this something that should be conceptually be public ? Because then I'll fix that the next time I update yosys on BCR

oharboe:
> is this something that should be conceptually be public ? Because then I'll fix that the next time I update yosys on BCR

One step at the time... First I need to get this to work, then caches must work(not invalidate hour/day long ORFS builds). Then cleanup like this...

oharboe:
@MrAMS Can we get rid of the verilog/chisel/verilator and use bazel bcr?

See CI errors...

PR #663, opened 2026-04-08
bump: update orfs, openroad, qt-bazel and fix stale patch

oharboe:
- orfs -> b6ff3eb63dac (drop patch 0001 merged upstream)
- openroad -> d34d035725c3
- qt-bazel -> 886104974c2f
- rules_shell 0.6.1 -> 0.7.1 (fix version mismatch warning)
- Regenerate patch 0035 for new upstream MODULE.bazel

### DIFF
--- a/chisel/openroad-remove-initrunfiles.patch
+++ b/chisel/openroad-remove-initrunfiles.patch
@@ -1,0 +1,140 @@
+diff --git a/bazel/BUILD b/bazel/BUILD
+index a15adee..66f99d2 100644
+--- a/bazel/BUILD
++++ b/bazel/BUILD
+@@ -124,18 +124,12 @@ cc_library(
+     }),
+ )
+ 
+-# shim old library as it is still referenced in src/sta, implementing the
+-# old behavior.
++# Stub library kept for src/sta:opensta compatibility.
++# Tcl setup is now handled by //bazel:tcl_library_init.
+ cc_library(
+     name = "runfiles",
+     srcs = ["InitRunFiles.cpp"],
+-    data = [":tcl_resources_dir"],
+     visibility = ["//visibility:public"],
+-    deps = [
+-        "@rules_cc//cc/runfiles",
+-        "@tcl_lang//:tcl",
+-    ],
+-    alwayslink = True,
+ )
+ 
+ # small build test to check if "bazel build //src/sta:StaTclInitVar" works
+diff --git a/bazel/InitRunFiles.cpp b/bazel/InitRunFiles.cpp
+index 0651774..ece50b0 100644
+--- a/bazel/InitRunFiles.cpp
++++ b/bazel/InitRunFiles.cpp
+@@ -1,86 +1,3 @@
+-// SPDX-License-Identifier: BSD-3-Clause
+-// Copyright (c) 2026, The OpenROAD Authors
+-
+-// This file to go away as soon as we use the tcl_library_init initialization
+-// in OpenSTA.
+-
+-#include <unistd.h>  // readlink()
+-
+-#include <climits>
+-#include <cstdlib>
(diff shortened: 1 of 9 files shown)


oharboe:
@hzeller :crossed_fingers:

oharboe:
@hzeller for now caching is mythical for my use-cases... lots of churn to update. Previously I would get a binary directly from ORFS. But previously I would have to wait for the patches to land upstream before I could even think about upgrading. Now I can upgrade and patch upstream and actually test the patches before I consider upstreaming them. Which is great.

But caching remains elusive...

hzeller:
If things don't work well with the cache, maybe build a tiny project in the CI and see how it behaves.

I [once set up a cache in OpenROAD](https://github.com/The-OpenROAD-Project/OpenROAD/commit/79994a8d4d0d38c997cd1a6ce9f7f5c0e7cc65bf) (while we were still on github actions), that also used time as factor, so you get the latest cache not something that matches alphabetically first with the hash (though not sure that it is still needed; that was needed around 2020 when I first started setting up bazel caches).

One problem can be if the cache size is more than 10GiB, it will exceed what github has to offer.

oharboe:
Disappointment is mismanaged expectations, could you check how big the cache needs to be and if there is anything we can do about it? I'd like to stay within the github budget.

oharboe:
@hzeller I want to see if we can whittle the testing setup down so that it fits into github budget. Perhaps we can do something on the openroad side?

oharboe:
@hzeller The [main build of this PR](https://github.com/The-OpenROAD-Project/bazel-orfs/actions/runs/24179101305) that I just merged. It *should* have caching. This PR ran in 45 minutes, this should run in much less than that...

hzeller:
that ci is still building. Look at the end when it packs up its caches if it runs into trouble
You should be able to inspect your caches in 
https://github.com/The-OpenROAD-Project/bazel-orfs/actions/caches

chisel/openroad-remove-initrunfiles.patch
```diff
@@ -0,0 +1,140 @@
+diff --git a/bazel/BUILD b/bazel/BUILD
```
oharboe, line 1: @hzeller @maliberty Have you run into this? 

Could I be running into this with the [static web files](https://github.com/The-OpenROAD-Project/OpenROAD/pull/10087#issuecomment-4212445227)?

hzeller: can you give more context ?

oharboe: See link above:

```
[Warning] Failed to create bazel runfiles: ERROR: external/rules_cc+/cc/runfiles/runfiles.cc(103): cannot find runfiles (argv0="/home/oyvind/OpenROAD-flow-scripts/tools/OpenROAD/tmp/test/orfs/gcd/_main/openroad")
application-specific initialization failed: 
```

oharboe: Have to run... so many things in flight, but I'm optimistic that all of these issues will trickle as fixes into OpenROAD and ORFS :-) Takes a bit of time though, step by step...

hzeller: Is this a recent openroad from after April 2nd ?

hzeller: also: I have no idea what this patch is attempting to accomplish

hzeller:
https://github.com/The-OpenROAD-Project/bazel-orfs/actions/runs/24179101305/job/70567318431#step:31:2 <- looks like it wants a path of sorts to save something ?
I also have no idea what this bazel setup thing is doing, I'd probably do the bazel install and cache storage manually to know exactly what is done.

PR #665, opened 2026-04-10
ci: replace setup-bazel repository cache with explicit actions/cache

oharboe:
setup-bazel's repository-cache option fails in its post step with "Path Validation Error: Path(s) specified in the action for caching do(es) not exist". Replace it with an explicit actions/cache step at a known path (~/.cache/bazel-repo), matching how disk cache is already handled. Both caches are now configured in ~/.bazelrc.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,14 @@ jobs:
         uses: bazel-contrib/setup-bazel@0.19.0
         with:
           bazelisk-cache: true
-          repository-cache: true
+
+      - name: Repository cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/bazel-repo
+          key: bazel-repo-${{ runner.os }}-${{ github.sha }}
+          restore-keys: |
+            bazel-repo-${{ runner.os }}-
 
       - name: Disk cache
         uses: actions/cache@v4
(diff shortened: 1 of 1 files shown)


oharboe:
@hzeller cache doesn't work yet. :crying_cat_face:

.github/workflows/ci.yml
```diff
@@ -50,8 +57,10 @@ jobs:
           restore-keys: |
             bazel-disk-${{ runner.os }}-
 
-      - name: Configure disk cache path
-        run: echo 'common --disk_cache=/home/runner/.cache/bazel-disk' >> ~/.bazelrc
+      - name: Configure caches
+        run: |
+          echo 'common --repository_cache=~/.cache/bazel-repo' >> ~/.bazelrc
+          echo 'common --disk_cache=~/.cache/bazel-disk' >> ~/.bazelrc
```
hzeller: you're adding paths here that you are not packing in the actions/cache part. Is this intentional ?

.github/workflows/ci.yml
hzeller, line 73: while investigating the cache, maybe build something really small (so, not `...`) so that you can see turnaround faster: unpacking cache, and later actually writing something to cache.

oharboe:
works, it seems :relieved:

.github/workflows/ci.yml
oharboe: it was running in the background while I was doing other stuff, but yes, horrible workflow if I had to wait for it...

.github/workflows/ci.yml
oharboe: Hmm... I didn't understand that...

PR #666, opened 2026-04-10
The long road to ditch docker

oharboe:
This directory is a downstream consumer of `bazel-orfs` that builds and
tests a SystemVerilog counter entirely from source — no docker, no
oss-cad-suite. It patches OpenROAD and ORFS where needed.

## Strategy

1. **Patch openroad/orfs here** to demonstrate everything works end-to-end.
2. **Fix bazel-orfs issues** in bazel-orfs itself (separate commits).
3. **Create minimum-churn PRs for OpenROAD** once patches are proven here.
4. **OpenROAD adopts latest bazel-orfs**, getting rid of docker.

## How to use

```bash
cd test/downstream
bazelisk build ...    # synthesis via orfs_flow (slang frontend)
bazelisk test ...     # verilator simulation via cc_test + gtest
```

### bazel-orfs fixes (can merge independently)

- **mock-klayout leaks to downstream** — made `dev_dependency`, klayout
  now truly optional (`extension.bzl`, `config.bzl`, `environment.bzl`).

### Patches applied here (to be upstreamed)

- **OpenROAD `isolate = True`** — `npm` `use_extension` uses experimental
  `isolate` flag; patched out (`openroad-remove-isolate.patch`).

- **OpenROAD STA `Iterator.hh`** — GCC 15 / clang 20 enforce
  `-Wtemplate-body`; added `typename` and `this->` qualifiers
  (`openroad-fix-sta-iterator-templates.patch`).

- **ORFS slang plugin path** — `synth_preamble.tcl` hardcodes
  `plugin -i slang`; patched to respect `SLANG_PLUGIN_PATH` env var
  (`orfs-slang-plugin-path.patch`).

### Workarounds in `.bazelrc`

- **C++20** — OpenROAD STA uses `std::bit_cast`, `std::format`,
  `.contains()`. Until bazel-orfs sets this via transition rules,
  `--cxxopt=-std=c++20` is needed.

### Non-BCR deps require root-module overrides

`orfs`, `openroad`, and `qt-bazel` are not on BCR. Downstream consumers
must provide `git_override` entries (root-module-only). Ideally bazel-orfs
would deliver these through its extension.

### Python 3.13 toolchain must be registered by root module

bazel-orfs pip deps are locked to Python 3.13. The root module must
register it as the default toolchain via `python.toolchain(is_default = True)`.

### yosys visibility and headers patches

BCR yosys has `package(default_visibility = ["//visibility:private"])`.
Downstream consumers must duplicate the `yosys-visibility.patch` and add
a `yosys-hdrs.patch` for plugin compilation (`single_version_override` is
root-only).

### yosys-slang not on BCR

The slang yosys plugin must be built from source. This test includes a
native Bazel build (`yosys_slang.BUILD.bazel`) that compiles slang, fmt,
and yosys-slang into `slang.so`. A `merge_yosys_share` rule combines the
BCR yosys share tree with the plugin.


oharboe:
@hzeller FYI

test/downstream/patches/openroad-remove-initrunfiles.patch
```diff
@@ -0,0 +1,140 @@
+diff --git a/bazel/BUILD b/bazel/BUILD
```
hzeller, line 1: this looks like a patch not needed as `bazel/runfiles` is not used anymore anyway
(and in #10095, this it is removed, so this patch would conflict)

test/downstream/patches/openroad-fix-sta-iterator-templates.patch
```diff
@@ -0,0 +1,84 @@
+diff --git a/src/sta/include/sta/Iterator.hh b/src/sta/include/sta/Iterator.hh
+index 83e799e..25f043d 100644
+--- a/src/sta/include/sta/Iterator.hh
++++ b/src/sta/include/sta/Iterator.hh
+@@ -48,7 +48,7 @@ public:
+     seq_(seq)
+   {
+     if (seq_)
+-      itr_ = seq_->begin();
++      this->itr_ = seq_->begin();
```
hzeller, line 10: I see the typename specification below, but why is `this->` needed ?

hzeller: Also, isn't this something you'd like to upstream to parallax ?

test/downstream/patches/openroad-remove-initrunfiles.patch
```diff
@@ -0,0 +1,140 @@
...
+ #else
+   using rules_cc::cc::runfiles::Runfiles;
++
++  // Clear inherited RUNFILES_* env vars: when OpenROAD is invoked by
++  // a build system that previously ran another Bazel binary (e.g. a
++  // Python wrapper), those variables point to the *other* binary's
++  // runfiles tree.  Runfiles::Create() checks env vars first, so it
++  // would resolve paths in the wrong tree.  Unsetting forces it to
++  // fall back to Tcl_GetNameOfExecutable(), which is always correct.
++  unsetenv("RUNFILES_DIR");
```
hzeller, line 135: can you show that this is a problem (i.e. write  unit test) and then add the unsetenv to provide a fix ?

test/downstream/patches/openroad-visibility.patch
```diff
@@ -0,0 +1,49 @@
+diff --git a/BUILD.bazel b/BUILD.bazel
+index 0556c28..3fb282c 100644
+--- a/BUILD.bazel
++++ b/BUILD.bazel
+@@ -219,7 +219,7 @@ cc_library(
```
hzeller, line 5: why do these need to be public, and if so , why not send a PR to openroad direcly ?

test/downstream/patches/yosys-hdrs.patch
```diff
@@ -0,0 +1,15 @@
...
+         "passes/cmds/sdc/graph-stubs.sdc:sdc/graph-stubs.sdc",
+     ],
+ )
++
++# Header-only target for building out-of-tree yosys plugins.
++cc_library(
++    name = "hdrs",
++    hdrs = glob(["kernel/*.h", "kernel/*.inc"]),
++    defines = COMMON_DEFINES,
++    includes = ["."],
```
hzeller, line 13: is this includes needed ?

test/downstream/patches/yosys-hdrs.patch
```diff
@@ -0,0 +1,15 @@
+--- a/BUILD.bazel
++++ b/BUILD.bazel
+@@ -691,3 +691,12 @@
+         "passes/cmds/sdc/graph-stubs.sdc:sdc/graph-stubs.sdc",
+     ],
+ )
++
++# Header-only target for building out-of-tree yosys plugins.
++cc_library(
```
hzeller, line 9: If this turns out to be useful, we should ask the BCR maintainer of yosys to add.

test/downstream/patches/yosys-visibility.patch
```diff
@@ -0,0 +1,11 @@
+--- a/BUILD.bazel
++++ b/BUILD.bazel
+@@ -1,5 +1,8 @@
++# Techlib cell definitions, plugins and other runtime data needed by the
++# synthesis flow.
+ share_tree(
+     name = "yosys_share",
++    visibility = ["//visibility:public"],
```
hzeller, line 8: Same here (bring to BCR)

hzeller (commented):
Had a quick look.

oharboe:
will address in followup PR

oharboe:
@hzeller Spooky, pull request to get rid of docker is #666

test/downstream/patches/openroad-fix-sta-iterator-templates.patch
oharboe: I prototype fixes here and then when I know what is needed I can create a PR

test/downstream/patches/openroad-visibility.patch
oharboe: I need fast turnaround prototyping and testing of patches in a full context. Only when I know what is needed will I articulate patches upstream.

test/downstream/patches/yosys-visibility.patch
oharboe: I still don't know exactly what is needed, I need to get a private repository upgraded not to use Docker before I have confirmation that I know exactly what is needed. test/downstream/ is a public repository that articulates in non-NDA format what the problem is. It is also a way to track the changes. I'll whittle away at test/downstream as fixes make it into upstream repositories.

PR #668, opened 2026-04-10
fix: make sby tool labels configurable, fix parallel synth SDC

oharboe:
Two fixes:

1. sby/sby.bzl: Make sby, yosys, yosys_abc attrs public and mandatory on the rule, with string defaults in the sby_test macro. This moves Label resolution from bazel-orfs's repo mapping (where @oss_cad_suite is not visible) to the caller's BUILD file context.

2. private/rules.bzl: Fix parallel synthesis stubbing 1_synth.sdc with an empty file. Move it to the ODB generation action outputs so synth_odb.tcl produces the real canonicalized SDC.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,6 +48,7 @@ jobs:
           key: bazel-repo-${{ runner.os }}-${{ github.sha }}
           restore-keys: |
             bazel-repo-${{ runner.os }}-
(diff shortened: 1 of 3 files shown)


oharboe:
@hzeller :crying_cat_face: Why is this so hard?

image

oharboe:
@hzeller Cache works again

PR #681, opened 2026-04-12
docs: reframe toolchains idea around caching and pinning

oharboe:
Rename from "Pain-Relief Toolchains and Packages" to "Caching and Pinning: Toolchains and Packages". Establish caching and pinning as complementary strategies: caching for application code, pinning for foundational tools where rebuilds are costly and often unnecessary.

Add multi-registry artifact hosting section covering public GitHub Releases, private GitHub Releases, GCR/Artifact Registry, and generic HTTPS backends. All use the same urls + sha256 mechanism with authentication handled by environment tokens or Bazel credential helpers.

Add build-pin-share workflow for users who need custom patches: build from source once, package, upload to any hosting backend, pin by sha256 in MODULE.bazel.

### DIFF
--- a/ideas/toolchains.md
+++ b/ideas/toolchains.md
@@ -1,17 +1,71 @@
-# Pain-Relief Toolchains and Packages
+# Caching and Pinning: Toolchains and Packages
 
 ## Context
 
-Two things make bazel-orfs painful for new users and CI:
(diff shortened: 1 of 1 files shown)


oharboe:
@maliberty @hzeller @MrAMS FYI, not pulling the trigger on this yet, but I can see this is needed.

Thoughts welcome!

PR #686, opened 2026-04-14
feat: extract yosys + plugins from ORFS image (yosys-latest)

mguthaus:
## Summary

Add a parallel lazy extraction of yosys from the ORFS Docker image, so downstream projects that need yosys plugins (notably `yosys-slang` for native SystemVerilog parsing via `SYNTH_HDL_FRONTEND=slang`) can get them without rebuilding yosys from source. The BCR ``@yosys`` package doesn't ship the plugins directory, so today there is no way to use ``@bazel-orfs//:openroad-latest`` alongside a plugin-enabled yosys without extracting it yourself.

## What's added

- **``docker_orfs_yosys_image``** — new repo in ``extension.bzl``, uses the existing ``docker_pkg`` rule with the same image/sha256 as ``docker_orfs_image``. Bazel's content-addressed repository cache dedupes layer downloads, so enabling both repos adds no net bandwidth.
- **``docker_yosys.BUILD.bazel``** — exposes three targets extracted from ``OpenROAD-flow-scripts/tools/install/yosys``: ``yosys``, ``yosys-abc``, and ``yosys_share`` (including ``plugins/slang.so``).
- **Three aliases in ``//:BUILD``** — ``yosys-latest``, ``yosys-abc-latest``, ``yosys-share-latest`` — mirroring the existing ``openroad-latest``/``opensta-latest`` public surface.
- **``test/docker/MODULE.bazel``** — demonstrates routing ``orfs.default(yosys=..., yosys_abc=..., yosys_share=...)`` at the new aliases.

## Design choice: separate repo, not extended ``docker_openroad.BUILD.bazel``

Kept the yosys extraction in its own repo (``docker_orfs_yosys_image``) rather than adding targets to ``docker_openroad.BUILD.bazel`` so that users who don't need image-sourced yosys pay nothing — the yosys share tree is extracted only when one of the new aliases is actually referenced.

## Test plan

- [x] ``test/docker`` workspace MODULE updated to use the new aliases
- [ ] CI green on the docker-test workspace

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -5,11 +5,16 @@ load("@rules_python//python:defs.bzl", "py_binary", "py_library", "py_test")
 load("@rules_python//python:pip.bzl", "compile_pip_requirements")
 load("@rules_shell//shell:sh_binary.bzl", "sh_binary")
 
-# OpenROAD and OpenSTA binaries from the latest ORFS Docker image.
-# Downstream projects use these labels to skip building OpenROAD from source:
+# OpenROAD, OpenSTA, and yosys binaries from the latest ORFS Docker image.
+# Downstream projects use these labels to skip building these tools from source:
 #   orfs.default(
 #       openroad = "@bazel-orfs//:openroad-latest",
 #       opensta = "@bazel-orfs//:opensta-latest",
+#       # Optional: when the design needs yosys plugins (e.g. slang) that
+#       # the BCR @yosys package doesn't ship.
+#       yosys = "@bazel-orfs//:yosys-latest",
+#       yosys_abc = "@bazel-orfs//:yosys-abc-latest",
+#       yosys_share = "@bazel-orfs//:yosys-share-latest",
 #   )
 # The Docker image is only downloaded when these targets are actually built.
 alias(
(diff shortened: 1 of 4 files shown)


oharboe:
we are sunsetting docker image usage in favour of bazel + patches + claude and possibly pinning.

@maliberty FYI

oharboe:
See alsl https://github.com/The-OpenROAD-Project/bazel-orfs/blob/main/ideas/toolchains.md

oharboe:
Please reopen if in light of this we should consider this PR

mguthaus:
Closing this — after reading `ideas/toolchains.md` more carefully, adding another `docker_pkg` extraction is swimming against the current direction. The doc explicitly says `@docker_orfs` / Docker-image extraction is being retired in favor of prebuilt yosys/OpenROAD toolchains hosted as GitHub Release assets, with the yosys prebuilt slated to bundle the slang plugin directly.

The right path for downstream users who need slang is to wait for the `bazel-orfs-yosys` toolchain submodule (with `yosys.prebuilt(version=...)` / source fallback) and then switch to it, rather than adding more extraction machinery that will need to be pruned later. Happy to help with the toolchain work if useful — let me know.

oharboe:
It is of course completely reasonable to set up in your own repo the use of a docker image openroad/yosys binary and point bazel-orfs to that binary and that is a use case bazel-orfs intend to support as a first class citizen.

Issue #690, opened 2026-04-18
Help wanted with upstreaming yosys patches


oharboe:
@hzeller Could you help upstreaming these?

Part of getting rid of docker cleanly...

- [ ] https://github.com/The-OpenROAD-Project/bazel-orfs/blob/main/patches/yosys-hdrs.patch
- [ ] https://github.com/The-OpenROAD-Project/bazel-orfs/blob/main/patches/yosys-visibility.patch
- [x] https://github.com/The-OpenROAD-Project/bazel-orfs/blob/main/patches/abc-main-c-as-cpp.patch

oharboe:
@maliberty @hzeller

Hit a Bzlmod papercut today worth filing away.

`bazel-orfs` uses `git_override` in its MODULE.bazel to pin a few dependencies that aren't in BCR. That works fine when you build `bazel-orfs` standalone. The moment you add it as a dep in another project, your MODULE.bazel becomes the root, and Bazel silently ignores every `*_override` from non-root modules. Resolution then either fails (dep isn't in any registry) or quietly drifts to a different version than `bazel-orfs` was tested against.

The workaround is for every downstream user to copy-paste the same `git_override` block into their own root MODULE.bazel and keep it in sync forever, with patches. Ouch. Duplicating `git_override` is somewhat tolerable when waiting for bazel bcr of every module, but patches hurt more I think.

The real fix is upstreaming: get the pinned deps into BCR so a normal `bazel_dep(..., version = ...)` resolves transitively without any override gymnastics. Until that happens, every consumer of `bazel-orfs` pays the tax, and we'll keep paying it on every other module that does the same thing.

Filing this under "reasons we should keep pushing patches upstream rather than carrying them." No action needed, just wanted it on your radar.

hzeller:
For abc, wouldn't https://registry.bazel.build/modules/abc work ? (in 0.64, there was some c -> cpp patch going on, maybe tha tis what you need ?)

oharboe:
should be fixed here https://github.com/The-OpenROAD-Project/bazel-orfs/pull/697

oharboe:
more yosys patches to retire...

oharboe:
adequately tracked in patches, no?

PR #693, opened 2026-04-19
fix: bump abc to 0.64-yosyshq and target sswPart.c for C++ compilation

oharboe:
hzeller pointed out in #690 that BCR abc 0.64-yosyshq already handles the C++-source-in-.c-file problem for src/misc/util/utilPth.c by renaming it to utilPth.cpp via copy_file. A .c src is compiled as C by rules_cc and falls back to host libstdc++ headers; the resulting objects reference std::__glibcxx_assert_fail and fail to link against the hermetic libc++. Renaming to .cpp picks up the C++ compile profile (and -stdlib=libc++) so the symbols come from libc++.

Bump abc from 0.62-yosyshq to 0.64-yosyshq to pick up that upstream fix, and replace our old main.c -> main.cc patch (which only forced the cc_binary link path and didn't actually fix the cc_library compile) with a patch that mirrors the 0.64 pattern for the one remaining <atomic> user: src/proof/ssw/sswPart.c.

A straight copy_file relocates sswPart into bazel-out/ which breaks its quoted sibling #include "sswInt.h" (quoted includes are resolved relative to the source file's directory). Also add src/proof/ssw to the cc_library's includes attribute so sswInt.h is reachable via the include search path from the relocated sswPart.cpp.

Verified @abc//:abc_bin builds from a clean cache and the resulting binary runs ABC commands.


oharboe:
@hzeller upgraded, old patch gone, new patch

hzeller:
ok , so there are a bunch of files that should possibly be compiled as c++. Maybe I send a PR for abc where these are provided as a list and renamed.

The main.c file now works (because that is the one you changed in the previous patch) ?

oharboe:
> ok , so there are a bunch of files that should possibly be compiled as c++. Maybe I send a PR for abc where these are provided as a list and renamed.

Help greatly appreciated! I've never interacted with the abc project.

> The main.c file now works (because that is the one you changed in the previous patch) ?

To be honest: Claude is doing this for me. I'm just "herding patches", I don't know why they are needed. There are some things I engage in others, I just observe that Claude thinks it is needed.

E.g. I've suggested to Matt that we mock-array tests could be rewritten to SystemVerilog to drop Chisel dependencies, so there I know more about what is going on.

PR #694, opened 2026-04-19
Orfs info arguments

oharboe:


test/auto_utilization.tcl
```diff
@@ -0,0 +1,21 @@
...
+set num_cells [llength [$block getInsts]]
+
+if {$num_cells < 500} {
+    set util 50
+} elseif {$num_cells < 5000} {
+    set util 40
+} else {
+    set util 30
+}
+set density [format "%.2f" [expr {$util / 100.0 + 0.15}]]
```
oharboe, line 14: @maliberty @MrAMS Just FYI, bazel-orfs now has an idiomatic way to compute arguments, such as CORE_UTILIZATION and PLACE_DENSITY based on the output of other bazel rules. This particular example is just a simple smoketest. This is a feature that comes in handy for more advanced flows and also area mocking for macros.

PR #696, opened 2026-04-20
chore: bump rules_chisel 0.3.0 -> 0.3.1, drop verilator cruft

oharboe:
rules_chisel 0.3.1 contains the VERILATOR_ROOT runfiles fix (hw-bzl/rules_chisel#1) and bumps verilator to 5.046.bcr.3 (hw-bzl/rules_chisel#2), so the local single_version_override patch that carried those same changes is no longer needed.

While here, drop unused bazel_dep entries that are now covered transitively by rules_chisel / rules_verilator:

- verilog/MODULE.bazel: rules_verilator, verilator (only rules_verilog is loaded here)
- chisel/MODULE.bazel: rules_verilog (unused)


oharboe:
@MrAMS Thanks!

PR #699, opened 2026-04-20
fix: build @gnumake hermetically with zig cc + musl

oharboe:
The host cc produced a host-dependent make binary whose content hash cascaded into Bazel remote-cache misses for every orfs_run / orfs_flow action (make is the Makefile-wrapper entrypoint for every stage).

Download Zig and bootstrap make with `zig cc` instead — statically linked against musl on Linux, native libc on macOS. With -ffile-prefix-map, -Wl,--build-id=none and -Wl,--strip-all the output is byte-identical across machines regardless of host glibc or host compiler version.

Verified: two builds in independent output_bases on this host produced identical sha256 1bc7ae3936cebe31eabf3b28686d8e6c56c76a7cdc2cc7dd0ad125930289d3d0 (320 KB, statically linked, stripped).

### DIFF
--- a/gnumake.bzl
+++ b/gnumake.bzl
@@ -1,36 +1,152 @@
-"""Repository rule that builds GNU Make from source."""
+"""Repository rule that builds GNU Make hermetically from source.
 
-_VERSION = "4.4.1"
-_SHA256 = "dd16fb1d67bfab79a72f5e8390735c49e3e8e70b4945a15ab1f81ddb78658fb3"
+The host `cc` would produce a host-dependent make binary whose content hash
+cascades into Bazel remote cache misses for every orfs_run / orfs_flow action
+(make is the Makefile-wrapper entrypoint for every stage).
+
+Instead, we download Zig and bootstrap with `zig cc`. On Linux we statically
+link against musl so the output is byte-identical across machines regardless
+of host glibc or host compiler version; on macOS we target the native libc
+(Apple doesn't allow fully-static libc linking).
+"""
+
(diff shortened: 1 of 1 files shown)


oharboe:
@hzeller FYI

oharboe:
@mguthaus Whittling away at hermiticity....

hzeller:
ahem, this downloads a huge zig compiler (gmake is written in C, not zig) (it takes > 10 minutes to download).

Why not just compile gmake with the regular c compiler ? With that, you get the hermeticity guarantees.

Whipping out a BUILD file shoud be simple. Then upstreaming it to BZR will also make it useful for others.

oharboe:
> ahem, this downloads a huge zig compiler (gmake is written in C, not zig) (it takes > 10 minutes to download).
> 
> Why not just compile gmake with the regular c compiler ? With that, you get the hermeticity guarantees.
> 
> Whipping out a BUILD file shoud be simple. Then upstreaming it to BZR will also make it useful for others.

How can this be done hermetically. This matters *ENORMOUSLY* otherwise we have hours and days of rebuild of ORFS flows.

oharboe:
I think I tried and it failed for reasons. Will have another go.

hzeller:
> How can this be done hermetically. 

Whatever you build using the same sources (gnumake) and toolchain (the llvm toolchain in the MODULE.bazel) with bazel _is_ hermetic. Compile it on two different machines, and you get a bit-identical output.

oharboe:
Should be fixed now. OpenROAD needs a version bump is all.

PR #700, opened 2026-04-21
build: stub libxml2.so.2 in the LLVM toolchain to drop host libxml2-dev

oharboe:
Every C++ link action in bazel-orfs — including the OpenROAD build — currently emits:

    ld.lld: /lib/x86_64-linux-gnu/libxml2.so.2: no version information
    available (required by .../ld.lld)

`ld.lld` from LLVM 20.1.8 dynamically links `libxml2.so.2`, but only calls into it for Windows manifest merging (`lld/COFF/DriverUtils.cpp`). On Linux the symbols are never touched, yet ld.so still has to resolve the library at load time — leaking a silent host dependency on `libxml2-dev` into an otherwise hermetic toolchain, and peppering every link with a noisy version-mismatch warning.

Fix: inject an abort-on-call stub `lib/libxml2.so.2` into the LLVM archive, where ld.lld's `RUNPATH=$ORIGIN/../lib` resolves it before the host's copy. The stub is compiled by the just-extracted `./bin/clang` with `-nostdlib -nostdinc -fuse-ld=lld`, so its bytes are independent of the host compiler — critical for remote-cache hits since the stub lands in every link action's sandbox.

The BCR `toolchains_llvm` bazel_dep has no `patch_cmds` hook, so we use the `llvm.toolchain_root` escape hatch: we declare our own `http_archive` for the LLVM tarball (same url/sha256 `toolchains_llvm` would have fetched) with `patch_cmds` that builds and drops the stub into `lib/`, then point `llvm.toolchain_root` at the patched repo. The BUILD file we give that repo is a copy of `@toolchains_llvm//toolchain:BUILD.llvm_repo` with `lib/libxml2.so.2` added to the `:lib` filegroup so it reaches link actions' sandboxes.

Scope: only helps when bazel-orfs or gallery is the root MODULE (toolchains_llvm is a `dev_dependency` so downstream consumers bring their own toolchain). Downstream consumers that want the same hermeticity can copy the pattern from `tools/xml-hack/` and the two MODULE.bazel blocks — or, ideally, see the upstream-fix checklist below, fix it properly, and delete both.

Where this should really be fixed — please nerd-snipe yourself:

 * llvm/llvm-project: `lld/COFF/DriverUtils.cpp` conditionally links libxml2. Guard it by target OS, or static-link the tiny manifest merger on Linux releases, and every downstream user gets the fix.
 * llvm/llvm-project release build: ship `ld.lld` statically linked on Linux release tarballs.
 * bazel-contrib/toolchains_llvm (BCR): bundle a libxml2 stub in the distribution, or add a `patches=` / `patch_cmds=` attr on the `llvm.toolchain` module-extension so nobody has to detour through `toolchain_root` to apply a one-file fix.

When this commit can be reverted: see `tools/xml-hack/README.md`. Any one of the upstream fixes above is sufficient; try deleting the hack, rebuild, and if the warning is gone and `ldd bin/ld.lld` shows no host libxml2 dependency, send the revert PR.

Verified locally:

 * `ldd external/.../llvm_toolchain_llvm/bin/ld.lld` resolves `libxml2.so.2` to the toolchain's own `lib/libxml2.so.2` (the stub), not `/lib/x86_64-linux-gnu/`.
 * `ld.lld --version` prints cleanly; no libxml2 warning.
 * `bazelisk aquery CppLink` on a test binary lists the stub as an input to the link action (sandboxed next to `ld.lld`).
 * `bazelisk build //tmp_link_test:test` under processwrapper-sandbox links with no warning.
 * `bazelisk run //:fix_lint` clean.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -81,6 +81,28 @@ bazel_dep(name = "glpk", version = "5.0.bcr.4")
 
 bazel_dep(name = "toolchains_llvm", version = "1.5.0", dev_dependency = True)
 
+# Stub libxml2.so.2 — boarding up a broken window.
+# ld.lld from the LLVM release links libxml2 dynamically but only calls into
+# it for Windows manifest merging (lld/COFF). On Linux we get a useless
+# "no version information available" warning on every link and a hidden
+# host libxml2-dev requirement. We inject an abort-on-call stub into the
+# toolchain's lib/, which ld.lld's RUNPATH=$ORIGIN/../lib resolves before
+# the system library. See tools/xml-hack/README.md for deletion conditions.
+# toolchains_llvm has no patch_cmds hook, so we bring our own LLVM tarball
+# via an http_archive and point llvm.toolchain_root at it.
+http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "llvm_toolchain_llvm",
+    build_file = "//tools/xml-hack:BUILD.llvm_repo",
+    patch_cmds = [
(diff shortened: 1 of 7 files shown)


oharboe:
@hzeller @mguthaus FYI

hzeller:
As I said earlier this looks like an unnecessary hack: usually updating the
llvm version will bring out one that does not need libxml

oharboe:
## LLVM upgrade does not drop the libxml2 hack — yet

Tested on Ubuntu 25.10, `bazel-orfs` main.

### What I tried

Branched from `origin/main`, bumped `toolchains_llvm` 1.5.0 → 1.7.0, bumped LLVM 20.1.8 → 21.1.8 (22.1.4 isn't in `toolchains_llvm` v1.7.0's distribution list), deleted `tools/xml-hack/` and the `http_archive` / `llvm.toolchain_root` blocks from both `MODULE.bazel`s, then built a trivial `cc_binary`.

### What happened

Same warning, verbatim:

```
external/toolchains_llvm++llvm+llvm_toolchain_llvm/bin/ld.lld: /lib/x86_64-linux-gnu/libxml2.so.2: no version information available (required by ld.lld)
```

Direct probe of the release tarballs with `readelf -d bin/lld`:

| LLVM | `NEEDED libxml2.so.2`? | Ships `lib/libxml2.so.2`? |
|---|---|---|
| 20.1.8 | yes | no |
| 21.1.8 | yes | no |
| 22.1.4 | yes | no |

Host side: Ubuntu 25.10 ships libxml2 **2.14.5** with SONAME `libxml2.so.16`. `/lib/x86_64-linux-gnu/libxml2.so.2` is a compat symlink to `libxml2.so.16` without the old `LIBXML2_2.4.30` / `LIBXML2_2.6.0` version tags — which is why the warning fires here specifically.

### The upstream fix exists, but got reverted

- [llvm/llvm-project#166867](https://github.com/llvm/llvm-project/pull/166867) "Add support for statically linking libxml2, enabled by default for releases" — merged 2026-04-10, explicitly motivated by `bazel-contrib/toolchains_llvm#657` (our exact issue).
- [llvm/llvm-project#191609](https://github.com/llvm/llvm-project/pull/191609) reverted it 2026-04-13 after breaking Fuchsia's Windows CI.
- LLVM 22.1.4 cut 2026-04-21 **without** the fix.

### Verdict

Keeping the hack. The README's deletion checklist (`tools/xml-hack/README.md`) is still the right trigger — once #166867 (or equivalent) re-lands and ships in a release tarball, delete the hack and send the revert.

PR #702, opened 2026-04-21
feat: emit per-macro pre-layout .lib for ideal-clock pre-CTS STA

oharboe:
@maliberty @tspyrou FYI, dual charcterization...

Hierarchical macro abstracts generated by `write_timing_model` after CTS bake the propagated clock insertion delay into dedicated `min/max_clock_tree_path` timing arcs on each clock input pin. When the parent consumes such a `.lib` during its own pre-CTS stages (synth/floorplan/place) the parent's STA over-budgets timing into the macro, penalizing synthesis and placement.

Dual characterization is the standard industry fix: run `generate_abstract` once at post-`place` to emit an ideal-clock variant (the `min/max_clock_tree_path` arcs are emitted with zero values because no clock tree exists yet), and once at `final` for the propagated variant. The parent's pre-CTS stages consume the ideal lib; CTS+ consume the propagated one. `.lef`/`.gds` stay single-canonical because pin locations commit at place and macro-internal OBS differences don't matter when macros are treated as black boxes by the parent router.

User-facing API is unchanged: a single `:<name>_generate_abstract` label still represents the macro on both sides. Internally:

- `orfs_flow` with `abstract_stage` past `place` now also emits `:<name>_pre_layout_generate_abstract` fed from the post-`place` .odb, and wires it into the canonical abstract via a new `pre_layout_abstract` attr on `orfs_abstract`.
- `OrfsInfo` gains `lib_pre_layout` and `additional_libs_pre_layout`; `orfs_additional_arguments` takes a `use_pre_layout` flag applied in synth/2_floorplan/3_place when building `ADDITIONAL_LIBS`.
- Fixes a latent same-file `mv` in `renames()` that's newly reachable when a pre-layout variant's src forwards base-variant files.

Verified on //test:L1MetadataArray_generate_abstract:
  tag_array_64x184/base/*_typ.lib:       max_clock_tree_path = 31.21806
  tag_array_64x184/pre_layout/*_typ.lib: max_clock_tree_path =  0.00000
  synth/floorplan/place args.mk → pre_layout lib
  cts args.mk                    → canonical lib

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -68,7 +68,11 @@ git_override(
     # qt-bazel upstream ships it as system_provided = True; we replace
     # that with a static cc_library compiled from xcb-util-cursor 0.1.5.
     patch_cmds = [
-        "curl -sL https://xorg.freedesktop.org/archive/individual/lib/xcb-util-cursor-0.1.5.tar.xz | tar xJ --strip-components=1 -C interface_libs/xcb xcb-util-cursor-0.1.5/cursor",
+        # Retry + --fail so transient xorg.freedesktop.org hiccups don't
+        # feed a truncated/HTML response into tar (seen in CI as "xz:
+        # File format not recognized"). Stage to a file first for a
+        # clearer error if extraction fails.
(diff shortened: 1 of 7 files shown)


oharboe:
@maliberty @tspyrou So dual charcterization should improve QoR for timing driven placement, no?

PR #707, opened 2026-04-23
feat: tools/memory_macro_scaler — DSE-grade memory .lib/.lef predictor

oharboe:
Adds a new public tool under //tools/memory_macro_scaler that predicts area, timing, and power for ASAP7 memory macros from a closed-form fit over published OpenRAM + DFFRAM numbers, and emits .lib + .lef pairs ready for consumption by orfs_macro(). Two entry points:

  scale_macro()        — rewrite a reference ORFS abstract to idiomatic
                         values; supports dual-input and single-input
                         (place-stage, no pre_layout sibling) modes.
                         Emits both OrfsInfo.lib and lib_pre_layout via
                         a new scaled_macro_lib custom rule.

  behavioral_macros()  — from behavioral Verilog, emit one orfs_macro()
                         per memory module without running synthesis or
                         P&R. The only EDA step in the flow is a light
                         Yosys read/elaborate to get authoritative pin
                         endpoints; everything else comes from the fit.

Fit is a two-parameter log-log regression per memory kind (sram / ff) with data points from OpenRAM FreePDK45 (Cornell ECE5745 anchor), OpenRAM sky130, and AUCOHL DFFRAM sky130. Multi-PDK via tech_nm² Dennard area scaling so the same fit predicts FreePDK45, sky130, and ASAP7. Access time calibrated to OpenRAM's 128×32×1RW FreePDK45 anchor with CACTI-shaped (log₂(rows)·√bits) scaling. Port factors from the OpenRAM paper. Power anchored to ~5 pJ/access at 45 nm and leakage to ~1 pW/bit at 45 nm. Published residuals: ±1% FF, ±25% SRAM; a unit test pins the budget.

Banking of oversized shapes follows the primer's rule-3 decomposition: word slice → row bank → read-port replicate → write-port address-bank, with per-bank limits 512 rows / 128 bits / 2R / 1W. Aggregation: area sums, access time picks up log₂(row_banks) FO4 of mux select, read/write energy scales with word_slices (parallel), leakage sums over every bank.

Characterization subsystem:
  characterization/asap7_sweep.yaml       committed, schema-documented
                                          sweep-results file; auto-loaded
                                          into the fit at import time.
  characterization/sweep_configs.py       shape grid spanning depth /
                                          width / port / bit-line-RC /
                                          masked-write regimes; each
                                          regime cited against DFFRAM /
                                          OpenRAM / RegFileStudy / CACTI.
  characterization/generate_sweep.py      per-.lib harvester.
  characterization/pin_asap7_sweep.py     bazel-run target that writes
                                          the YAML back into the source
                                          tree via BUILD_WORKSPACE_DIRECTORY.

Tests: 42 inline-fixture Python unit tests plus 6 Bazel-level tests (lib_pre_layout_test on the scaled macro, file accumulation, dual-char content diff, place-only variants); full suite runs in seconds.

gallery/megaboom wiring: MEMORY_BLACKBOXES now has a side-by-side behavioral_macros() invocation tagged manual, emitting :<module>_behavioral targets for comparison against the existing demo_sram() flow.

Both .bzl entry points are listed in EXTERNAL_API.md as public surface.

### DIFF
--- a/EXTERNAL_API.md
+++ b/EXTERNAL_API.md
@@ -19,6 +19,8 @@ consumers.
 | `generate.bzl` | `fir_library` | none (`@circt` http_archive) |
 | `orfs_genrule.bzl` | `orfs_genrule` | none |
 | `sweep.bzl` | `sweep` | none (via openroad.bzl) |
+| `tools/memory_macro_scaler/scale_macro.bzl` | `scale_macro` — dual-characterization macro scaler wiring | none (via openroad.bzl) |
+| `tools/memory_macro_scaler/scaled_macro_lib.bzl` | `scaled_macro_lib` — OrfsInfo-carrying wrapper around two scaled .lib files | none |
 
 ## Required non-dev bazel_dep entries
 
--- a/gallery/megaboom/BUILD.bazel
+++ b/gallery/megaboom/BUILD.bazel
@@ -3,6 +3,7 @@ load("//:gallery.bzl", "demo_gallery_image")
 load("@bazel-orfs-verilog//:generate.bzl", "fir_library")
 load("@bazel-orfs-verilog//:verilog.bzl", "verilog_directory", "verilog_single_file_library")
 load("@bazel-orfs//:openroad.bzl", "orfs_flow")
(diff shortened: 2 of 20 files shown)


oharboe:
@ovebryne FYI

PR #708, opened 2026-04-23
feat: memory_macro_scaler — distinguish sync vs async read in generat…

oharboe:
…ed .lib

The generated Q-pin arc is the contract STA consumes: a sync-read memory needs `timing_type : rising_edge` (CLK→Q) so STA treats Q as a flip-flop output, while an async-read memory needs `timing_type : combinational` (ADDR→Q) so STA treats Q as a mux-style propagation delay. Previously generate_lib() hard-coded `rising_edge` for every output pin, so flop memories and any async-read SRAMs were silently modeled as zero-delay flip-flops — wrong kind of arc, wrong results from downstream STA.

- MemoryRole gains a `read_mode` field; flop memories default to "async".
- generate_lib() branches on read_mode: async emits combinational ADDR→Q and skips setup/hold on pure-R addr/en pins. RW ports keep their write-side setup/hold since the address pin is shared with the synchronous write function.
- scan_verilog_for_memories() classifies read mode heuristically: an NBA to the rdata pin (direct or via a registered stage) under a posedge- always is sync; `assign rdata = <array>[...]` is async; ambiguous defaults to sync.
- Tests pin the Q-arc shape for sync, async, RW, and the end-to-end Verilog→.lib path; scale_lib_text() now has a regression test that confirms a combinational timing_type isn't silently rewritten.


oharboe:
@MrAMS memory macro handling. If you have behavioral model RAM on asap7, you can simply specify that they should be represented as macros and you get something that should be good 'nuf for DSE and idiomatic for asap7.

PR #717, opened 2026-05-07
gnumake: drop zig bootstrap, build with cc_binary

oharboe:
PR #699 made @gnumake hermetic by downloading the full Zig toolchain (~80 MB, hzeller measured >10 min) and bootstrapping make with zig cc -static -target x86_64-linux-musl. hzeller's review:

  ahem, this downloads a huge zig compiler (gmake is written in C, not
  zig) (it takes > 10 minutes to download). Why not just compile gmake
  with the regular c compiler? With that, you get the hermeticity
  guarantees. Whipping out a BUILD file shoud be simple. Then upstreaming
  it to BZR will also make it useful for others.

He's right that the project already registers a hermetic clang (@llvm_toolchain from toolchains_llvm, used to build OpenROAD/yosys), so zig is redundant for fetching a C compiler. Convert @gnumake to a thin repo rule that downloads the gmake source and overlays a checked-in cc_binary BUILD file plus a vendored config.h.

Hermeticity for the make binary itself comes from:
  - Bazel sandbox -> execroot-relative paths in __FILE__ (no host paths leak into the binary).
  - linkopts = ["-Wl,--build-id=none", "-Wl,-s"] strip the random ELF build-id and symbols/debug info.
  - Same hermetic clang the rest of the project uses.

Verified two clean output_bases on this host produce identical sha256 c678430fbbd1cff14b8b22e67080187fa49c762fbc7d45b8481a71e5d34770a9 (290 KB, stripped, dynamically linked). `make --version` reports GNU Make 4.4.1. A trivial Makefile recipe runs end-to-end.

What this does not attempt: vendoring a sysroot to make the binary byte-identical across host glibc versions. cc_binary outputs are currently tied to host glibc symbol versions -- the same caveat that already applies to cc_binary-built yosys/openroad in this project. Project-wide cross-distro hermeticity is a separate, larger effort.

Linux x86_64 only -- the registered LLVM tarball
(LLVM-20.1.8-Linux-X64.tar.xz) doesn't support other hosts today, so the previous gnumake.bzl support for macOS/aarch64 was already dead.

config.h is the autoconf output of `./configure --without-guile --disable-nls` against gmake 4.4.1, captured once and committed verbatim. HAVE_DECL_BSD_SIGNAL is patched to 0 because bsd_signal was removed from glibc 2.27+; the original config.h captured under musl had it set to 1.

### DIFF
--- a/gnumake.bzl
+++ b/gnumake.bzl
@@ -15,6 +15,52 @@ yosys/openroad equally and should be a project-wide change.
 
 Linux x86_64 only — that's the only host the registered LLVM tarball
 supports today.
+
+Substituting a host make
+------------------------
+
+On hosts where the registered LLVM toolchain doesn't apply (macOS,
+non-x86_64 Linux), or when a developer simply wants to use their own
+make, the standard Bazel escape hatch is `--override_repository`. It
+replaces the entire `@gnumake` repo with a local directory — the
+`gnumake()` repo rule's tarball download is skipped, no compile is
+attempted, no bazel-orfs-specific helper is involved.
+
+One-time setup (any path of the user's choosing):
+
+    mkdir -p ~/.config/bazel/host_make
+    touch ~/.config/bazel/host_make/REPO.bazel
+    cat > ~/.config/bazel/host_make/make.sh <<'EOF'
+    #!/bin/sh
+    exec /opt/homebrew/bin/gmake "$@"
+    EOF
+    chmod +x ~/.config/bazel/host_make/make.sh
+    cat > ~/.config/bazel/host_make/BUILD.bazel <<'EOF'
+    sh_binary(
+        name = "make",
+        srcs = ["make.sh"],
+        visibility = ["//visibility:public"],
+    )
+    EOF
+
+Then in `user.bazelrc` (gitignored, per-machine):
+
+    build --override_repository=+orfs_repositories+gnumake=/Users/foo/.config/bazel/host_make
+
+The override key is the *canonical* repo name `+orfs_repositories+gnumake`
+(bzlmod adds the `+orfs_repositories+` prefix because @gnumake is created
+by the orfs_repositories module extension). The apparent name `gnumake`
+will not match — Bazel silently keeps using the bazel-built make if you
+write that.
+
+The `REPO.bazel` file is just a Bazel marker that says "this directory
+is a repo root"; it can stay empty.
+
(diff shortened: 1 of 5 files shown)


oharboe:
@hzeller This is what Claude did...

tools/gnumake/BUILD.gnumake
```diff
@@ -0,0 +1,69 @@
...
+    ] + glob([
+        "src/*.h",
+        "lib/*.h",
+    ]),
+    copts = [
+        "-DHAVE_CONFIG_H",
+        # gmake's remake.c / read.c reference these as install-prefix paths
+        # (.LIBPATTERNS, dlopen search). Match autotools' default --prefix
+        # so behavior is identical to a stock distro build.
+        '-DLIBDIR=\\"/usr/local/lib\\"',
```
hzeller: thes paths don't really make sense. What happens if you set them empty or not at all ?

gnumake.bzl
```diff
@@ -1,103 +1,26 @@
...
+cross-distro byte identity (the previous zig+musl-static guarantee) would
+require vendoring a sysroot and is out of scope here; it would benefit
+yosys/openroad equally and should be a project-wide change.
+
+Linux x86_64 only — that's the only host the registered LLVM tarball
+supports today.
 """
 
 _MAKE_VERSION = "4.4.1"
 _MAKE_SHA256 = "dd16fb1d67bfab79a72f5e8390735c49e3e8e70b4945a15ab1f81ddb78658fb3"
```
hzeller: where is this hash used to verify the download ?

oharboe:
Pushed two follow-up commits:

  * 65454e8 — addresses both inline review comments (inlined the sha256 hash; LIBDIR/LOCALEDIR/INCLUDEDIR
  -D values now empty strings rather than fictional /usr/local paths).
  * 61f4d64 — docs only, adds an `--override_repository` recipe for users who want to swap
  @gnumake//:make for a host gmake (e.g. on macOS). No bazel-orfs helper, just stock Bazel.

  Verified end-to-end here: `bazelisk test //...` is 47/47 green both with the default and with the
  override active against /usr/bin/make.

oharboe:
@hzeller OK now?

oharboe:
@hzeller Ok?

hzeller:
the extension stuff still looks a bit complicated, but maybe that is what needed with bazel 8 ? Anyway, I think this is a good start; from there we should get it on bzr, then we don't need any *.bzl files.

PR #725, opened 2026-05-12
rules: fix runtime paths for orfs_run_executable env vars

oharboe:
## Summary

Follow-up to #724. The merged \`orfs_run_executable\` rule emits a wrapper that doesn't set \`OPENROAD_EXE\` / \`OPENSTA_EXE\` / \`KLAYOUT_CMD\` / \`FLOW_HOME\`, so the orfs Makefile falls back to its hardcoded \`external/orfs+/tools/install/OpenROAD/bin/openroad\` path — which doesn't exist in a runfiles tree (orfs+ runfiles only include \`flow/\`, not \`tools/install/\`).

## Failure mode

When the rule's emitted wrapper invokes make's \`run\` target, the recipe tries:

\`\`\`
.../external/orfs+/tools/install/OpenROAD/bin/openroad -exit -no_init -threads 48 ...
bash: line 1: openroad: No such file or directory
make: *** [.../flow/Makefile:846: run] Error 127
\`\`\`

## Fix

Convert each tool's \`.short_path\` (which uses \`../<repo>/...\` for external repos in Bazel 7+) into \`external/<repo>/...\` form and pass them as make variable overrides. Combined with the wrapper's existing \`external -> \$(pwd)/..\` symlink, this gives cwd-invariant runfiles paths — important because the orfs Makefile cds into a results dir before invoking openroad.

Also switch the \`make\` and \`makefile\` paths in the wrapper to use the same \`_runtime_path\` helper so they're consistent (previously \`make\` used \`.short_path\` directly while \`makefile\` used \`.path\`).

## Verified

End-to-end: an actual TCL drill-down target at a downstream consumer (ascenium) running \`bazelisk run //hardware/tile:wns-families-run -- TO_GLOB='...' MAX_ITERATIONS=40\` against a cached synth ODB. Validates that openroad / opensta / klayout / Makefile are all reachable at runtime.

## Test plan

- [x] \`bazelisk run //:fix_lint\` — formatted
- [x] \`bazelisk build //:run_executable\` — py_binary builds
- [x] End-to-end run against a real ORFS synth target reaches openroad (no path resolution failure)

### DIFF
--- a/private/rules.bzl
+++ b/private/rules.bzl
@@ -646,12 +646,50 @@ def _run_executable_impl(ctx):
     else:
         work_home = None
 
+    # Convert a `.short_path` (which uses `../<repo>/...` for external repos
+    # in Bazel 7+) into an `external/<repo>/...` form. Combined with the
+    # `external -> $(pwd)/..` symlink the wrapper sets up at runtime, this
+    # gives us cwd-invariant paths into the runfiles tree — important when
+    # the orfs Makefile cds into a results dir before invoking openroad.
+    def _runtime_path(short_path):
+        if short_path.startswith("../"):
+            return "external/" + short_path[3:]
+        return short_path
+
+    # Workspace-relative paths live at `_main/<short_path>` in the runfiles
+    # tree (where `_main` is the main workspace's repo dir), so we anchor
+    # them off `$RUNFILES/_main`. External-repo paths (`../<repo>/...` form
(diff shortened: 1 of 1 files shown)


oharboe:
CI failure inherited from main — `@@openroad+//src/sta:opensta` analysis error affects every PR rebased on current main (run 25738413665 on main itself also fails with the same error). Not caused by this change; will re-test once main's CI is green.

PR #752, opened 2026-07-02
Add yosys_prefix_tar packaging macro and :yosys_tar target

oharboe:
Adds a `yosys_prefix_tar()` macro (rules_pkg, already a dep) and a
plugin-free `//:yosys_tar` instantiation that package the hermetic
`@yosys` BCR build as a relocatable PREFIX tarball:

```
bin/yosys
bin/yosys-abc
share/yosys/...
share/yosys/plugins/*.so   # optional, via plugins = [...]
```

Yosys resolves its data directory as `<bindir>/../share/yosys/` and abc
as `<bindir>/yosys-abc`, and the bazel build links
readline/tcl/zlib/ffi/abc statically, so the tarball extracts into any
prefix (e.g. `/usr/local`, `~/.local`, `tools/install/yosys`) without
patching.

Motivation: give installers a single source of truth for yosys — the
`@yosys` pin in bazel-orfs — instead of ad-hoc git-clone-and-make
builds. First consumer is ORFS `etc/DependencyInstaller.sh` /
`bazel/install.sh` (The-OpenROAD-Project/OpenROAD-flow-scripts#4325); ORFS vendors this
change as a `git_override` patch in the meantime, so this PR can land
whenever convenient.

Consumers that want plugins bundled instantiate the macro themselves,
e.g.:

```starlark
load("@bazel-orfs//:yosys_pkg.bzl", "yosys_prefix_tar")

yosys_prefix_tar(
    name = "yosys_tar",
    plugins = ["@yosys-slang//src/yosys_plugin:slang.so"],
)
```

Known gap: the tarball does not include the `make install` extras
(`yosys-config`, `yosys-smtbmc`, `yosys-witness`, `yosys-filterlib`);
those can be added later if a consumer needs them.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -4,6 +4,7 @@
 load("@rules_python//python:defs.bzl", "py_binary", "py_library", "py_test")
 load("@rules_python//python:pip.bzl", "compile_pip_requirements")
 load("@rules_shell//shell:sh_binary.bzl", "sh_binary")
+load(":yosys_pkg.bzl", "yosys_prefix_tar")
 
 exports_files([
     "bump.py",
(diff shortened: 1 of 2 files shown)


oharboe:
@maliberty Concept phase, in case someone pulls on the rope to use Bazel instead of DependencyInstaller.sh to build yosys

PR #754, opened 2026-07-06
synth: single writer for 1_synth.sdc, fail hard on missing outputs

oharboe:
Root-cause fix for the read-only `1_synth.sdc` failure hzeller hit on RBE (OpenROAD PR [#10818](https://github.com/The-OpenROAD-Project/OpenROAD/pull/10818) worked around it with `chmod +w`).

## The bug

`1_synth.sdc` had two writers in ORFS:

1. yosys' `synth.tcl` copied the user's `SDC_FILE` into place — a leftover from before SDC canonicalization existed ("One day a more sophisticated synthesis will write out a modified .sdc").
2. `synth_odb.tcl` then overwrote it **in place** with the OpenSTA-canonicalized version (`write_sdc -no_timestamp`).

`cp` preserves mode bits, so with read-only inputs (bazel RBE) the copy is read-only and the canonicalization write fails. On writable local checkouts it only accidentally worked.

## The fix

**patches/0037** (carried until the identical fix lands in ORFS upstream, then dropped at the next bump):
- `synth.tcl` / `synth_preamble.tcl` no longer produce `1_synth.sdc`; the yosys stage's SDC output is `1_2_yosys.sdc` (raw copy), as before.
- `synth_odb.tcl` / `synth_syn.tcl` are the sole writers of `1_synth.sdc`, and write `1_synth.odb`/`.sdc` unconditionally: producing them is the `do-1_synth` contract. The `WRITE_ODB_AND_SDC_EACH_STAGE=0` no-op silently left downstream stages consuming the stale raw copy.

**rules.bzl** (fail hard and early instead of festering):
- `1_synth.sdc` is declared together with `1_synth.odb` only when `save_odb=True`, instead of as a yosys output. (`save_odb=False` users only consume the `1_2_yosys.v` output group.)
- Stop touch-masking primary artifacts (`1_2_yosys.v`/`.sdc`, `1_synth.odb`/`.sdc`) after make: a synth step that silently produced nothing used to yield empty files that flowed downstream — and would let idempotency tests compare two empty files as "identical". Now bazel fails the action with "declared output was not created". Logs, reports and `mem.json` remain touch-stubbed since the `SYNTH_NETLIST_FILES` path legitimately skips them.

## Testing

- `bazelisk test //test/...`: 43 pass; 17 fail to build locally due to the pre-existing `@scip` glibc 2.41 compile issue (unrelated — real-OpenROAD targets; covered by CI).
- Verified the patch applies to the pinned ORFS and that the patched `@orfs` scripts contain a single `1_synth.sdc` writer.

Follow-ups after merge: `bazelisk run @bazel-orfs//:bump` in OpenROAD (unblocks the RBE tests, replaces #10818), and the same fix as an ORFS PR so patch 0037 can be dropped.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -42,6 +42,7 @@ archive_override(
     patch_strip = 1,
     patches = [
         "//patches:0036-fix-bazel-orfs-bazel_dep-non-dev-for-load-visibility.patch",
+        "//patches:0037-orfs-single-writer-1_synth-sdc.patch",
     ],
     strip_prefix = "OpenROAD-flow-scripts-fb65d5b487d2afd443fc1892330f622c776a57ac",
     urls = ["https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/archive/fb65d5b487d2afd443fc1892330f622c776a57ac.tar.gz"],
--- a/patches/0037-orfs-single-writer-1_synth-sdc.patch
+++ b/patches/0037-orfs-single-writer-1_synth-sdc.patch
@@ -1,0 +1,100 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?=C3=98yvind=20Harboe?= <oyvind.harboe@zylin.com>
+Date: Mon, 6 Jul 2026 09:58:32 +0200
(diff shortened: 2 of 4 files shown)


oharboe:
@hzeller FYI

PR #761, opened 2026-07-08
fix: load cc_binary from @rules_cc in gnumake BUILD overlay

maliberty:
## Problem

Bazel 9 removed the native \`cc_binary\` rule, so a BUILD file that calls it without a \`load()\` statement fails to load. The \`gnumake\` repo's BUILD overlay (\`tools/gnumake/BUILD.gnumake\`, added in #717 when gnumake switched from the zig bootstrap to a \`cc_binary\` build) uses \`cc_binary\` directly. This is fine on Bazel 8 (bazel-orfs's own \`.bazelversion\`), but breaks any consumer that reaches \`@gnumake//:make\` under Bazel 9 — e.g. OpenROAD-as-root, which is migrating to Bazel 9:

\`\`\`
ERROR: .../gnumake/BUILD.bazel:7:10: cc_binary(
  ...
  This rule has been removed from Bazel. Please add a \`load()\` statement for it.
ERROR: no such target '@@...+gnumake//:make': target 'make' not declared in package ''
\`\`\`

## Fix

- Load \`cc_binary\` from \`@rules_cc//cc:cc_binary.bzl\` in the overlay.
- Declare \`rules_cc\` as a \`bazel_dep\` so \`@rules_cc\` resolves in the generated repo's mapping (bazel-orfs did not previously depend on it directly).

Backward-compatible: works on both Bazel 8 and Bazel 9. Mirrors the mock_klayout \`sh_binary\` fix (#759).

## Test

- \`bazelisk build --nobuild @gnumake//:make\` resolves on bazel-orfs (Bazel 8.6.0).
- With this change carried as a patch, OpenROAD-as-root passes \`bazelisk test //test/...\` 45/45 under Bazel 9.1.1.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -15,6 +15,7 @@ local_path_override(
 bazel_dep(name = "buildifier_prebuilt", version = "6.4.0", dev_dependency = True)
 
 bazel_dep(name = "rules_pkg", version = "1.2.0")
+bazel_dep(name = "rules_cc", version = "0.2.18")
 bazel_dep(name = "rules_shell", version = "0.6.1")
 
 bazel_dep(name = "mock-klayout", version = "0.0.1", dev_dependency = True)
--- a/tools/gnumake/BUILD.gnumake
+++ b/tools/gnumake/BUILD.gnumake
@@ -4,6 +4,8 @@
 # am__append_6 (remote-stub), and the lib/Makefile am_libgnu_a_OBJECTS
 # from a configure run on Linux glibc. See tools/gnumake/config.h for the
 # matching #define set.
+load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
+
 cc_binary(
     name = "make",
     srcs = [


oharboe:
I don't need to review PRs here of this kind, I'm very glad you are engaing directly as a maintainer and merging.

A note on stability/breakage:

1. I find that it is very rare that we see any breakage with changes in bazel-orfs, can't recall. I've asked Claude to make a survey of this risk.
2. All users of bazel-orfs pin to a specific commit, so even if main is broken, it is trivial to carry a revert locally or a future PR. One can pin to an unmerged PR even.

oharboe:
## Breakage & rework in bazel-orfs — last 6 months

Reviewed the 506 commits from 2026-01-09 to 2026-05-23 for things we had to walk
back, restore, or fix a second time. Goal: gauge breakage risk and where reviews
pay off.

### By the numbers

| | count | % of commits |
|---|---|---|
| Total commits | 506 | — |
| `fix:` commits | 117 | 23% |
| bump / chore | 77 | 15% |
| touches a patch | 46 | 9% |

A ~23% fix rate is a high-churn, high-rework period — driven by a large
infra re-platforming (docker image → source-built tools, BCR migration for
yosys/abc/tcl/libffi/chisel, rules_python 1.6.3 → 1.8.5, gnumake source build,
OpenROAD pinning rework). **Every walk-back in the window traces to that
migration.** There is essentially no logic-bug revert thrash in this window.

### What we actually walked back or fixed twice

1. **rules_python 1.6.3 → 1.8.5** — upgrade + PYTHONPATH bootstrap fix shipped
   together (`dea261e`, `05737ec`, 03-03), then needed a second pass 10 days
   later (`58451e5` "fix pythonwrapper brittleness", 03-13).
2. **rules_cc / rules_shell upgrade — fully reverted** (`e317319`, 03-19):
   broke downstream WORKSPACE consumers.
3. **gnumake build strategy — reversed**: zig cc + musl (`6226c2e`, 04-21) →
   dropped for plain cc_binary (`943cae5`, 05-10).
4. **deps tarball packaging — same-day re-fix**: `00afb97` (04-06 10:28) →
   `33f2d89` "restore pkg_tar-based packaging from main" (04-06 17:32).
5. **Carried upstream patches — chronic**: unsetenv RUNFILES patch restored
   twice (`151150d`, `02b08f9`); patch 0035 restored then dropped
   (`811d761` → `0184aa0`); 0036 context re-refreshed twice; 0037 added then
   dropped next day (`964d545` → `a6d7dc1`); OpenROAD pin tarball staging wrong
   location fixed in same push (`40759ce` → `d18bf13`, "not /tmp").
6. **mock runfiles paths** (docker→source-tools migration): fixed, then fixed
   *again* (`62876a5` → `1cc11d1`), then glob'd (`4cc59e6`).

### Risk assessment

- **Risk is concentrated and predictable**, not spread across feature work.
  It lives in `MODULE.bazel`, carried patches, and toolchain/build-strategy
  changes. Feature commits (orfs_flow args, memory_macro_scaler, render_gds,
  kept_macros) show no walk-backs.
- **The failures are not visible in a diff**: WORKSPACE cycles, Python bootstrap
  semantics, patch-context drift, `/tmp`-vs-workdir staging, downstream breakage.
  Line-by-line review of the changed file catches almost none of these — they
  only surface when a full flow or a submodule/downstream build runs.
- **Patch maintenance is structurally fragile**: patches silently drop on every
  ORFS/OpenROAD rebase and get re-restored. This will keep recurring until they
  are upstreamed.

### Recommendation

Blanket code review has limited value here; **verification depth** does. Two
gates would have caught the majority of the in-window rework:

1. **Full-flow + submodule/downstream build gate** on any PR touching pins,
   patches, or `MODULE.bazel` — not just a green build of the changed file.
2. **"Did any carried patch silently drop?" check** on every bump.

Reserve human review for **reverts, "remove/drop X" commits, and build-strategy
changes**, where confident-but-wrong decisions and weeks-later reversals happen.

maliberty:
> I don't need to review PRs here of this kind

Ok I didn't want to upset your production flow but I'm fine to just merge.

PR #762, opened 2026-07-12
docs: /openroad-debug command for diagnosing OpenROAD/ORFS failures

oharboe:
Adds a `/openroad-debug` slash command capturing the reusable OpenROAD/ORFS **failure-diagnosis** knowledge that sits *before* packaging a reproducer, and indexes the debugging commands in `CLAUDE.md` as the single source of truth.

## What's new

`.claude/commands/openroad-debug.md` — the diagnosis layer that complements the existing packaging commands:

- **Decode the exit code** — openroad never self-terminates; 143/137/134/139 and how to find *who* sent the SIGTERM.
- **Characterize a hang** before calling it a deadlock — `futex`/`wchan`/`perf` sampling of the real process, the `-threads 1` race-vs-cost test, the gdb Heisenbug caveat.
- **Fast local iteration on an OpenROAD *source* change** — `_deps` + bring-your-own-binary, so an openroad edit re-runs one stage (minutes) instead of a full P&R (hours).
- **Split a stage `.tcl` at an ODB checkpoint** — re-run only a read-only late step; with the RSZ-0074/GRT-0265 caveat for repair-after-GRT.
- **Reproducer shape** — self-contained `repro.tcl` + a short `run.sh` that takes the openroad path (bring-your-own binary; strip the pinned binary from `_deps` bundles).
- **whittle nuances** for intermittent/race bugs (why a pure race won't whittle) — cross-references `/openroad-issue` step 4 for the standard flow rather than duplicating it.
- **OpenSTA-from-netlist caveat** — `write_verilog`→`read_verilog` can normalize away an ODB-backed trigger.

`CLAUDE.md` gains a **Debugging OpenROAD/ORFS failures** section indexing `/openroad-debug`, `/openroad-issue`, `/untar-and-run-report`, `/odb-to-cpp` as the SOT, so downstream consumers can point at these instead of duplicating the mechanics.

All examples use neutral `//<project>:<module>_<stage>` naming. Confidentiality purge run over the outbound content: no local paths, usernames, employer/private names, private URLs, or embargoed material.


oharboe:
@hzeller @maliberty I've published some debugging procedures I had lying around. Guess who's going to be reading these...

PR #763, opened 2026-07-12
slang: adopt BCR sv-elab plugin; bump ORFS+OpenROAD to master

oharboe:
## Migrate the slang frontend to BCR `sv-elab`; bump ORFS + OpenROAD to master

`povik/yosys-slang` was renamed to `povik/sv-elab` and published on the Bazel
Central Registry (`sv-elab 2026-07-07`, slang-11 based), exposing the yosys
plugin at `@sv-elab//src/yosys_plugin:slang.so`. This adopts it and bumps the
from-source EDA stack to master to pick it up.

### Bump (root + gallery + test/downstream)
- ORFS → `c90beac`, OpenROAD → `c724b7da` (`origin/master` via
  `bump --head=openroad`); qt-bazel → `541ebf6`.
- **Retire patch 0036** — ORFS upstreamed the non-dev `bazel_dep(bazel-orfs)`
  fix (its `MODULE.bazel` now carries it with the same rationale).
- Sync direct-dep drift from OpenROAD master: `rules_cc` 0.2.20,
  `abc` 0.64-yosyshq.bcr.2, `glpk` 5.0.bcr.5; update bump.py's yosys/abc
  lockstep map to bcr.2.

### Adopt the BCR `sv-elab` slang plugin
- **root + gallery**: `bazel_dep(sv-elab)` +
  `orfs.default(yosys_plugins = ["@sv-elab//src/yosys_plugin:slang.so"])`;
  re-enable `//slang:test` (resolves the "slang.so is not in BCR yosys" TODO).
- **test/downstream**: point `merge_yosys_share` + `SLANG_PLUGIN_PATH` at
  `@sv-elab//src/yosys_plugin:slang.so`; **delete** the from-source
  `yosys_slang.bzl` + `yosys_slang.BUILD.bazel` (~200 lines); refresh README.

### bump.py cleanup
Retire the yosys-slang commit tracking (ORFS_TOOLS entry, `YOSYS_SLANG_MIN_COMMIT`
floor + compare-status gating, `update_yosys_slang_commit()`). The
`OPENROAD_SUBMODULES` `third-party/slang-elab` entry and the `@slang` sed stay —
OpenROAD still vendors that submodule for its in-tree `src/syn` elaborator.

### Validation
- OpenROAD master compiles from source; `//slang:test_synth` synthesizes cleanly
  with the slang-11 sv-elab plugin.
- `bump_test` green (111 tests).

### Yak-shaving — identified, not done
[...]

PR #764, opened 2026-07-12
docs: add debugging tips (synthesis, bumps, toolchain, overrides)

oharboe:
## docs: debugging tips

Adds `docs/debugging.md` capturing generalizable bazel-orfs debugging tips
learned while migrating to BCR `sv-elab` (#763) and diagnosing the slang-11
mock-array regression (OpenROAD #10884):

- **Synthesis / yosys** — unmask swallowed errors with `YOSYS_FLAGS=` (hidden
  even at the default `-v 3`); replay `read_slang` standalone; per-module
  canonicalization; the two independent slang consumers.
- **Builds** — "up-to-date, 0 processes" is a cache hit, not a compile; pipeline
  exit-code masking; hermetic toolchain registration is per-root.
- **Bumps** — `--head=openroad`; master-bump stale-patch drift; direct-dep
  drift; `--lockfile_mode=off`.
- **Overrides** — `patches` run before `patch_cmds`; local overrides skip
  `patch_cmds`; a module override must keep the module's declared name;
  extension-tag labels need non-dev deps.

Linked from README. Docs-only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,7 +287,9 @@ use_repo(orfs, "gnumake")
 ```
 
 First build takes 30-60 minutes; subsequent builds are incremental.
-See [docs/openroad.md](docs/openroad.md) for details and gotchas.
+See [docs/openroad.md](docs/openroad.md) for details and gotchas, and
+[docs/debugging.md](docs/debugging.md) for debugging tips (synthesis, bumps,
+from-source toolchain, overrides).
 
 #### Use locally installed tools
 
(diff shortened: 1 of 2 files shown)


oharboe:
Note on scope: the lead `YOSYS_FLAGS=` tip here is a workaround for an upstream
error-swallowing bug (slang errors routed through plain yosys `log()`, which
`-q`/`-v` detach from the console) — filed as
https://github.com/povik/sv-elab/issues/364. Once that lands, slang errors show
at the default `-v 3` and this tip stops being the way to see them.

Tracked in #765: demote it to a narrow "yosys `-q`/`-v` can still mask other
tools' `log()` output" note once the fix is in; the other tips are independent
and stand. (Also decided against a bespoke mock-array regression test — with
errors propagating it would be redundant.)

PR #766, opened 2026-07-12
sv-elab follow-ups: slang blackbox regression test + hermetic test/downstream build

oharboe:
Two follow-ups that harden the now-merged BCR `sv-elab` migration (#763). Both
validated locally.

### `slang`: blackboxed-module regression test
CI exercised the slang frontend (`//slang:test`) but never a **blackboxed
module** — the exact path that regressed in the sv-elab slang-11 migration
(OpenROAD #10884, where `--ignore-unknown-modules` was rejected). Adds
`//slang:blackbox_synth`: a top instantiating an empty `leaf` blackboxed by name
via `SYNTH_BLACKBOXES` + `--empty-blackboxes` through the slang frontend
(canonicalize + synth only; `save_odb = False`). A future sv-elab bump that
breaks blackboxing now fails **here at bump time** instead of surfacing in a
downstream. Verified: builds green.

### `test/downstream`: build OpenROAD from source with hermetic LLVM
`test/downstream` registered no C++ toolchain, so its from-source OpenROAD build
fell back to the host compiler and broke on newer glibc (`@scip`/`tinycthread`
on glibc 2.41) — it couldn't build locally and didn't faithfully model a real
downstream. Registers the zero-sysroot BCR `llvm` toolchain and mirrors the
hermetic-llvm compatibility overrides from the root `MODULE.bazel` (`sed`
simplified to drop the host-era memchr `patch_cmds`; add `bison`, `boost.icl`,
`gawk`, `m4`, `tcl_lang`). `counter_synth` now builds OpenROAD hermetically like
root/gallery, validating the BCR `sv-elab` plugin end-to-end. Verified: builds
green. README documents the requirement.

### DIFF
--- a/slang/BUILD
+++ b/slang/BUILD
@@ -1,4 +1,4 @@
-load("//:openroad.bzl", "orfs_flow")
+load("//:openroad.bzl", "orfs_flow", "orfs_synth")
 
 # The slang yosys plugin (slang.so) is provided from BCR sv-elab via
 # orfs.default(yosys_plugins=...) in MODULE.bazel.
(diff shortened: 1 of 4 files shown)


oharboe:
@hzeller FYI

PR #784, opened 2026-08-16
Implement orfs_tuner to extract ORFS stage as a standalone executable

oharboe:
This PR implements a `mock_tuner_executable` and `tuner.py.tpl` to extract ORFS stages as standalone standalone executables, along with a python mock optuna test, to enable bypassing Bazel for AI iterations.


oharboe:
This PR is significantly less brittle than the current `tuner_env.py` approach. Bypassing Make and using a Bazel template to bake the environment directly into the Python script (`ENV_JSON`) is much cleaner.

However, to fully support AI-driven and parallel DSE workflows, here is a wishlist of features that would be highly beneficial:

1. **Fix Concurrency Bug (Critical)**
   *Issue*: The script writes the wrapper to a hardcoded path: `wrapper_tcl_path = os.path.join(tmpdir, "tuner_wrapper.tcl")`. If multiple configurations run in parallel, they will overwrite each other's TCL wrappers.
   *Request*: Use Python's `tempfile.mkstemp` or `tempfile.NamedTemporaryFile` to generate unique wrapper files for each run.

2. **Add a `--keep` / `--debug` Flag**
   *Issue*: The `tuner_wrapper.tcl` is placed in `/tmp` and we might want to inspect it if a run fails.
   *Request*: Add an option to keep the generated wrapper files and output a distinct execution directory, similar to how we need `--keep` in our current workflow to inspect failures.

3. **Capture stdout/stderr to a File**
   *Issue*: The subprocess call directly inherits the parent's stdout/stderr.
   *Request*: Add a `--log-file` argument. When running 50 configurations in parallel via Optuna, having each standalone executable route its output to a specific log file makes debugging much easier than managing file handles in the parent process.

4. **Return Metrics JSON to stdout**
   *Issue*: The PR tells OpenROAD to write `-metrics tmpdir/metrics.json` but doesn't print it.
   *Request*: The script should read the generated `metrics.json` and print it to `stdout` upon completion. This restores the contract that the tuner executable yields a machine-readable JSON result.

oharboe:
I see Oyvind has already posted a great list of improvements. To fully support our macro-placement screening loops (`orfs_tune`), we also need the ability to inject Tcl structural overrides dynamically:

5. **Dynamic Tcl Injection (`--source` or `--tcl-command`)**
   *Issue*: The script currently only allows overriding environment variables (`--variable`). For structural experiments (like macro placement sweeps), we need to inject `set_inst_property` commands after the initial ODB is read but before the main flow stages.
   *Request*: Add a `--source <file.tcl>` argument that appends `source <file.tcl>` to the generated `tuner_wrapper.tcl` right before the main script is called.

PR #797, opened 2026-08-19
feat: explicit orfs_run() sources and dependency model

oharboe:
This PR introduces an explicit dependency and sources model for orfs_run().

### Changes:
* **Explicit dependencies**: Removes automatic harvesting of full-flow variables and configuration. Instead of passing magically gathered state, dependencies are declared explicitly via the sources parameter, matching orfs_flow().
* **Test Fixture (Estimation Ladder)**: Adds a robust estimation ladder test targeting a pipelined multiplier on the asap7 platform.
* **Shape Validation**: Includes a trend validation script (check_trends.py) wrapped in a build_test that ensures the achievable minimum clock period of the reg2reg path group logically degrades as the flow moves from synthesis to physical implementation.


oharboe:
Closing in favor of clean PR: the core sources API and stage variable filtering features landed in main via commits 2d7e53a and df84e7a. The estimation ladder test suite and Optuna tuning study are being submitted as a clean dedicated PR on top of latest main.

PR #801, opened 2026-08-22
bazel: vendor xcb-util-cursor in qt-bazel patch to eliminate flaky network fetches

oharboe:
## Summary

`openroad-qt` is on a deprecation track and Qt is not expected to change much.

Previously, `patch_cmds` in `MODULE.bazel` used `curl` to fetch `xcb-util-cursor-0.1.5.tar.xz` from `xorg.freedesktop.org`, which is notoriously flaky in CI and causes build timeouts.

This PR:
- Vendors the `xcb-util-cursor` source files directly inside `patches/qt-bazel-xcb-cursor-from-source.patch`.
- Removes `patch_cmds` and `curl` invocations from `git_override(module_name = "qt-bazel")` in `MODULE.bazel`.
- Completely eliminates network dependencies during the `qt-bazel` override patch phase.


oharboe:
@maliberty FYI - no action required

PR #823, opened 2026-08-24
feat(flow)!: drop the --define/settings variable-override channel

oharboe:
Removes the two Bazel-command-line channels for overriding ORFS variables:

- the `settings=` attribute (`BuildSettingInfo` build settings) threaded through `orfs_flow` and every stage rule, and
- the `ctx.var` (`--define`) per-stage override in `config_overrides()`, including the `VERILOG_TOP_PARAMS` hotfix that was FIXME'd for removal since ORFS PR 3746 landed.

Both are unused and were never tested or (in the `--define` case) documented. Overriding a variable from the Bazel command line invalidates every stage that reads it, which makes these channels a poor fit for design space exploration; the `orfs_arguments`/`extra_arguments` JSON overlay and `orfs_run_executable` have replaced them.

With `BuildSettingInfo` gone, `bazel_skylib` is only used by test BUILD files, so it is demoted to a `dev_dependency` per the EXTERNAL_API.md policy — one less MVS constraint imposed on downstream consumers.

Verified with `bazel query //...`, the mock sweep/stage-filter/parser/guard test targets, and `//:fix_lint`.

### DIFF
--- a/EXTERNAL_API.md
+++ b/EXTERNAL_API.md
@@ -12,7 +12,7 @@ consumers.
 
 | File | Exports | External deps |
 |---|---|---|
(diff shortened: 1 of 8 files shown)


oharboe:
@MrAMS Pruning dead-end features w.r.t. larget DSE/Optuna studies. orfs_run_executable() is *much* more scalable.

PR #826, opened 2026-08-25
fork/join snapshot idiom and orfs_run output directories

oharboe:
## Why

Configuration studies (like the `test/estimation_ladder` estimator sweeps) walk a *tree* of flow decisions, but every configuration re-pays the shared prefix — design load, floorplan, common placement stages — as its own OpenROAD process. POSIX `fork()` turns the running process into a copy-on-write snapshot: one process walks the tree depth-first, and every tree edge is computed exactly once. It is an in-memory `write_db`/`read_db` checkpoint that costs nothing to take and nothing to resume.

## What

One foreach-shaped Tcl command:

```tcl
source $::env(ORFS_FORK_TCL)

fork ?-parallel? ?-timeout seconds? varName valueList body
```

Sequential by default (fork is a snapshot mechanism, not parallelism: one active run owns the machine). Implicit join; returns a dict of per-value exit statuses. A crashed child loses only its subtree; a child that outlives `-timeout` self-destructs via its own `alarm(2)` (status 142) — enforced child-side because a parent-side kill would orphan running descendants that hold the machine and the stdout pipe open.

- **No OpenROAD patch.** The raw primitives live in `//fork:liborfsfork.so`, a Tcl-stubs extension compiled against the same `@tcl_lang` the openroad module embeds (MVS keeps them in lockstep), `load`ed into the unmodified binary.
- **Thread-pool safety baked in**: OpenSTA's dispatch workers (spawned by `-threads N`) die at fork and deadlock the first child that dispatches (observed as a `futex_wait` hang in timing-driven `global_placement`). `fork` quiesces the pool to one thread in the parent before forking — joining *live* workers, fully defined — so children run those paths inline, and restores after the join. OpenMP survives fork fine (validated empirically).
- **`orfs_run` gains `out_dir`**: a declared directory output whose path reaches the script as `$RUN_OUTPUT_DIR`, for scripts whose output file set is only known at runtime (one JSON per tree leaf). `orfs_run_executable` documents the same contract as a per-invocation absolute path, like `LOG_DIR`.
- Fixes a latent mock-openroad bug found en route: `puts $chan` ignored channels from `open`, so every scripted file write produced an empty file.

## Tests

- `//fork:fork_test` — the idiom's semantics in hermetic tclsh (1.1s): sequencing, statuses, crash tolerance, per-child timeout, nesting, `-parallel` barrier.
- `//test:run_out_dir_test` — `out_dir`/`$RUN_OUTPUT_DIR` end to end on mock-openroad (CI-fast).
- `//test:fork_smoke_test` (manual) — nested walk inside real OpenROAD after `load_design`, timing queries in forked children, a deliberate failing leaf, one JSON per leaf.

See `docs/fork.md` for the contract and the fork-hazard list. The follow-up PR rewrites the estimation_ladder studies on top of this (measured 3.7× on its test).

### DIFF
--- a/EXTERNAL_API.md
+++ b/EXTERNAL_API.md
@@ -28,6 +28,12 @@ These are loaded by the public `.bzl` files above or by BUILD files that
 downstream consumers may transitively evaluate:
 
 - `rules_shell` — `sh_binary` in ppa.bzl, root BUILD
+- `rules_cc` — `cc_binary` in fork/BUILD.bazel (the fork/join Tcl extension,
+  a default attr of `orfs_run`/`orfs_run_executable`)
+- `tcl_lang` — `tclstub` dependency of fork/BUILD.bazel; pinned to the same
+  version the openroad module embeds so the loadable extension always
+  matches the interpreter
+- `platforms` — `target_compatible_with` in fork/BUILD.bazel
 - `rules_verilog` — verilog providers in verilog.bzl
 - `rules_verilator` — verilator toolchain, verilator_cc_library
 - `verilator` — required by rules_verilator
@@ -47,7 +53,6 @@ These are only used by dev-only extensions, toolchains, or BUILD files:
 - `rules_java` — transitive dep of rules_scala
 - `rules_scala` — scala_config/scala_deps extensions
 - `rules_chisel` — chisel/test.bzl, dev BUILD files
-- `rules_cc` — chisel/test.bzl, dev BUILD files
 
 ## PDK extensibility
 
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -6,6 +6,9 @@ module(
 )
 
 bazel_dep(name = "bazel_skylib", version = "1.9.0", dev_dependency = True)
+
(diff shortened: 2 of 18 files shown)


oharboe:
@maliberty @MrAMS Toying with this to speed up DSE especially for hour long designs.

The idea is to get *much* faster design space exploration as what is to be explored is frequently a tree with measurements at leaves and we can explore the tree using fork/join without redoing more expensive higher up steps(like don't re-run global placement for each leaf) and no make or bazel involved w.r.t. dependencies.

If you're parameterizing the RTL, then this infrastructure doesn't work. In our case we redo Chisel to RTL, but we also do studies purely on the ORFS flow.

For something like autotuner, a single .tcl file flow, it can be a LOT faster. How much depends a lot on how expensive the first steps are and how far down in the flow you have to drive to get a good signal.

Something like autotuner is probably best done as a multi-rung study where the first rungs quickly scans a large population with few steps and low accuracy, then  prune the parameter space while keeping a good spread and ultimately doing only a few runs for the full flow for validation.

PR #846, opened 2026-08-26
feat(stages): FLOW_INPUT_VARIANT replaces cross-variant input renaming

oharboe:
A stage writes to its own variant's directories and reads its `src`'s. ORFS had one variant naming both, so every stage action began by `mv`-ing its inputs from the src variant's path to its own — string surgery on the last `/<src_variant>/` path segment, plus a second, wider pass (`rename_data`) for `generate_metadata`.

That renaming had leaked into six layers: `_remap`/`renames`/`input_commands`/`_mv_cmds`, the `rename_data` flag, `OrfsDepInfo.renames` threaded into the `_deps` tar manifest, a `for rename in $renames` loop in `deploy.tpl`, and the `renamed_inputs` attr on nine rules.

## The ORFS side

`patches/0038` splits the two directions:

```make
export FLOW_INPUT_VARIANT ?= $(FLOW_VARIANT)
export INPUT_RESULTS_DIR  ?= $(WORK_HOME)/results/$(PLATFORM)/$(DESIGN_NICKNAME)/$(FLOW_INPUT_VARIANT)
# ... and INPUT_LOG_DIR / INPUT_OBJECTS_DIR / INPUT_REPORTS_DIR
```

Defaulting to `FLOW_VARIANT` makes each `INPUT_*_DIR` expand to a string identical to its `*_DIR` counterpart, so an unchanged ORFS run is unchanged.

Reads resolve against a search path — `RESULTS_DIR` first, `INPUT_RESULTS_DIR` second. That ordering is required, not cosmetic: `do-place` runs 3_1…3_5 in one action, writing `3_1_place_gp_skip_io.odb` and reading it back, so a file this run just wrote must win over the upstream variant's copy.

`load_design` turned out to be the chokepoint — all 18 stage scripts hand it a bare basename, so routing it through `orfs_input_path` covers the whole flow. The rest is the `AUTO_MEMORIES` globs, `find_sdc_file`, the abstract's `.spef`, `yosys_load`'s netlist, and genMetrics' directories.

`find_sdc_file`'s candidate ordering now compares basenames rather than full paths. With one directory the two are equivalent (every candidate shares a prefix); with two, the directory prefix would otherwise decide which `.sdc` wins.

The Makefile's prerequisite graph is deliberately untouched — bazel-orfs drives `do-` targets.

## The bazel-orfs side

`renames()` collapses to one line beside the existing `FLOW_VARIANT`:

```python
"FLOW_INPUT_VARIANT": ctx.attr.src[OrfsInfo].variant,
```

Five commits, least-churn-first, each landing green:

1. `refactor(attrs): remove renamed_inputs` — unused in-tree, and `TESTING.md` listed it as never tested. If we want it back, the idiomatic shape is an `orfs_macro()`-style rule assembling an `OrfsInfo`, not a basename-keyed dict of file substitutions on every stage.
2. `feat(orfs): INPUT_*_DIR and FLOW_INPUT_VARIANT` — the patch alone, a no-op by construction.
3. `feat(stages): FLOW_INPUT_VARIANT replaces cross-variant input renaming` — the `mv` goes.
4. `refactor(deps): the _deps tarball stops renaming its inputs`.
5. `test(variants): pin down what replaced the cross-variant rename`.

Net: 189 lines deleted, 542 added — 407 of which are the patch, so bazel-orfs itself is a ~135-line net deletion.

## What proves it

After commit 3, **every existing test runs through the new path** — `FLOW_INPUT_VARIANT` is set unconditionally and the `mv` is gone. Full suite green (121/121).

Commit 5 adds three tests on variant "3" of the `lb_32x128` sweep, which forks at cts off base's `3_place.odb`: its `4_cts.mk` names `base` as input and `3` as output and the two have not collapsed; the action's inputs carry base's `3_place.odb` and no copy under the fork's own path; and the forked stages build under `bazel test` rather than only a wildcard build. All three fail with the `FLOW_INPUT_VARIANT` assignment removed.

`test/downstream` is a root module, so its `orfs` override repeats the patch list. It was carrying **no patches at all**, silently dropping 0037 too; it now mirrors the root.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -49,6 +49,7 @@ archive_override(
     patch_strip = 1,
     patches = [
         "//patches:0037-orfs-single-writer-1_synth-sdc.patch",
+        "//patches:0038-orfs-flow-input-variant.patch",
         "//patches:0039-orfs-slang-plugin-fallback.patch",
     ],
     strip_prefix = "OpenROAD-flow-scripts-427bd762b7b7448f8bb6bc4e14207aa3963fca30",
--- a/TESTING.md
+++ b/TESTING.md
@@ -98,7 +98,6 @@ comm -23 \
 | Feature | Notes |
 |---------|-------|
 | `stage_data` | Parameter exists in `orfs_flow`, never exercised |
-| `renamed_inputs` | Exists in flow.bzl/sweep.bzl, never tested |
 | `dissolve` in sweep | Feature exists (sweep.bzl), never used |
 | `orfs_update` | Rule exported, no test |
 | `save_odb=False` | Synthesis attribute, never tested |
--- a/deploy.tpl
+++ b/deploy.tpl
@@ -13,7 +13,6 @@ main() {
   local progname
   local config
   local genfiles
-  local renames
   local make
   local name="${NAME}"
   local package="${PACKAGE}"
@@ -32,11 +31,6 @@ main() {
       ;;
       -g|--genfiles)
         genfiles="$2"
-        shift
-        shift
-      ;;
-      -r|--renames)
-        renames="$2"
         shift
         shift
       ;;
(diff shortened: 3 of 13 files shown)


oharboe:
**Blocked on upstream, deliberately** — we do not want to carry `patches/0038` indefinitely.

Filed as The-OpenROAD-Project/OpenROAD-flow-scripts#4477. Once that merges and we bump the ORFS pin, this PR drops the patch file and the `MODULE.bazel` entry and becomes bazel-orfs changes only.

Rebased onto `main` (with #844 and #845 in). Full suite green, 123/123.

#847 no longer depends on this — it is standalone on `main` now.

oharboe:
First ORFS resolution, then revive this one

PR #853, opened 2026-08-27
bump: a 30-day rolling window, and delete what falls outside it

oharboe:
## Why

`bazelisk run //:bump` had grown two problems that share a root cause.

**It was slow.** Every run downloaded four multi-MB tarballs (ORFS,
OpenROAD, and OpenROAD's `src/sta`, `third-party/abc` and
`third-party/slang-elab` submodules) purely to hash them — serially, and
even when the resolved sha equalled the sha already in `MODULE.bazel`.

**It carried migration paths for shapes nobody writes any more.**
`bump.py` is a thin loader that downloads the newest `bump_impl.py` from
`main` on every run, so the *bumper* is never stale. The only thing that
can be out of date is the consumer's `MODULE.bazel` — and supporting
every shape bazel-orfs has ever emitted is what made the bumper large,
slow and risky.

## What

A 30-day rolling window, in both directions.

**For consumers.** `check_pin_age()` dates the commit that
`git_override(module_name = "bazel-orfs")` pins — one API call, before any
tarball is fetched — and hard-stops when it is more than 30 days old, or
when GitHub cannot date it at all. That pin dates the *shape* of the file,
since every other override in a consumer's `MODULE.bazel` was written by
whichever `bump_impl.py` that bazel-orfs shipped. The error names the
three ways out: re-seed from the README template, step the pin forward
≤30 days at a time, or `--allow-stale-pin` (unsupported, best-effort).
bazel-orfs itself has no such pin and is exempt.

**For the bumper's own compatibility code.** A migration branch is
introduced with a `# COMPAT(YYYY-MM-DD)` marker naming the date the old
shape stopped being written, and `//:bump_compat_test` fails once a marker
is older than the window. The fix is to delete the code the marker guards
— never to re-date the marker.

**Speed.** A block already pointing at the target commit has digests that
describe the same tarball, so they are reused (for openroad, read back out
of the generated `patch_cmds`; the block is still regenerated in case
patches moved). What is genuinely new is hashed in a thread pool. A
same-day re-bump now touches no tarball at all.

**Cruft removed.** The dead submodule-injection machinery (two empty
dicts and ~110 lines of no-ops), the `git_override(orfs)` fallback, and the
`git_override(openroad, init_submodules = True)` → `archive_override`
conversion. Both call sites now `_expect` an `archive_override` and say so;
an unrecognized block is left untouched under `--ignore` rather than half
rewritten.

## Commits

Least-churn-first, one concern each:

1. `refactor(bump)` — delete the dead submodule-injection machinery
2. `perf(bump)` — skip re-hashing unchanged pins, fetch archives concurrently
3. `feat(bump)` — hard-stop on a bazel-orfs pin older than 30 days
4. `refactor(bump)` — drop the out-of-window override migrations
5. `test(bump)` — fail the build on compat code that outlived the window

## Docs

- `docs/openroad.md` — "Supported window" and "Cleanup policy" (the
  source of truth)
- `README.md`, `docs/debugging.md` — cross-references
- `CLAUDE.md` — the agent-facing version: fix the file, don't route
  around the check
- `TESTING.md` — the new target

## Testing

```
bazelisk test //:bump_compat_test //:bump_impl_test //test/bump:bump_test
```

124 tests pass, up from 111. New coverage: the window gate (fresh, edge,
past-edge, undatable, exempt, variable-bound pin), the re-hash skip
(a second bump must fetch nothing), digest round-tripping, and that the
out-of-window shapes are refused rather than half-rewritten. Fixtures move
to the shapes bazel-orfs actually emits today.


oharboe:
Re-anchored both halves of the policy to commit dates — wall clock is no longer an input (`d042db0`).

**The pin gate.** `check_pin_age` becomes `check_pin_window(content, target_commit)`: it measures the span between the committer date of the pinned `bazel-orfs` commit and that of the commit the bump targets. `main()` resolves that target once, gates on it, and hands it to `bump()` so the commit that was gated is the commit that gets written. The same pair of commits always yields the same verdict.

**The cleanup test.** `//:bump_compat_test` no longer calls `date.today()`. It measures `COMPAT(YYYY-MM-DD)` markers against `bump_reference_date.txt` — the commit date of the ORFS pin, recorded by `//:bump` on a bazel-orfs self-bump. A given tree gets a fixed verdict, so re-running an old build cannot turn it red and a red build cannot be waited out. It also means compatibility code ages as the dependency stack advances rather than as the calendar turns.

128 + 10 tests pass, including one that evaluates the same commit pair with 1999 dates and expects the same span.

PR #860, opened 2026-08-29
test(estimation_ladder): a quantified PR verdict, and the fork machinery for it

oharboe:
Adds infrastructure and measurements for answering whether a change helped, on a
design where a single flow run cannot: on `multiplier_top` the achieved period
moves **25%** under perturbations that change nothing about the design. A nudge of
one site (0.054um, 0.014% of a 392um core) relocates all 16 macros by 144um on
average. So an infinitely fast flow would answer no better — latency was never the
binding constraint, variance is.

### Where the rationale lives

Deliberately in the files rather than in this description, since that is where
someone will look for it later:

| | |
|---|---|
| `test/estimation_ladder/README.md` | the findings, what they cost, and how to calibrate on another design |
| `docs/fork.md`, `fork/fork.tcl` | `-jobs` bounded scheduling, and why a forked child must stay single-threaded |
| `test/estimation_ladder/ci_gate.py` | why a verdict needs two bars, not one |

### What is added

- **`fork -jobs N`** — a bounded worker pool. `-parallel` forks every child at
  once, which put 42 OpenROAD processes on 16 cores; it is left alone rather than
  quietly bounded, because its unboundedness is a guarantee that
  `//fork:fork_test` relies on. The two are mutually exclusive.
- **Per-stage perturbation knobs** in the estimator, keyed so the batch walk
  diverges exactly where the perturbation takes effect.
- **`ci_gate`** — a PR verdict as a rank-based KPI (Hodges–Lehmann shift plus a
  win probability), always printed with the difference the ensemble could
  actually resolve.
- **`method_validation`** — the prerequisite that sets the accuracy floor by
  comparing against real flow ensembles.
- **`macro_stability`, `k_scaling`, `gate_cost`** — is the placer chaotic, does an
  ensemble buy resolution, and what does a member cost.
- **Fuzzable RTL twins** with known true effects, so a measured effect can be
  judged against one. `multiplier.sv` and `multiplier_top.sv` keep their published
  netlists.

### Tests

`//fork:fork_test` (adds a `-jobs` bound case and rejects `-parallel` with
`-jobs`), `//test:fork_smoke_test` (probes the thread-raising deadlock with a
deadline so it reports rather than hangs), and
`//test/estimation_ladder:seed_sensitivity_test` (the null control).
`//test/estimation_ladder:optuna_study_test` passes unchanged — the estimator
refactor is a no-op for existing studies.

### Limits worth knowing

- The method is validated on **one design and two variants, and it disagrees with
  the flow on one of them** — the small one. It catches a large regression and is
  blind near its resolution limit. The failure direction is the tolerable one: it
  returns inconclusive rather than a confident wrong answer.
- The accuracy floor (1.0%) is bracketed only by 0.45% and 9.43% from those two
  variants. It is design-specific; `method_validation` is how to re-measure it,
  and the README says so.
- `multiplier_top` has no flow reference: its 25% spread would need roughly 20+
  flow runs per variant to resolve the effect sizes in question.


oharboe:
Updated with the `multiplier_top` flow reference, which was the gap this
description previously listed as a limit. It changes a conclusion, so it is worth
calling out rather than leaving to a diff.

**The estimator's failure mode inverts on the macro design, and it inverts the
wrong way.** On `multiplier` it under-claims — returns inconclusive where the flow
sees an effect, which is safe. On `multiplier_top` it *over*-claims:

| variant | flow | estimator | agree |
|---|---|---|---|
| `roworder` | +1.52% [−3.73, +6.59] inconclusive | +9.72% [+3.80, +14.95] **worse** | no |
| `stage1` | −1.06% inconclusive | −0.45% inconclusive | yes |
| `stage4` | +19.68% **worse** | +37.78% **worse** | yes |

It calls `roworder` a confident ~10% regression where the flow cannot detect a
change at all. A false alarm is the failure that ends a KPI, because the first
developer to check one by hand finds nothing there. So the accuracy floor for this
design is **at least 10%**, against 1% on `multiplier` — the same "magnitudes do
not transfer" warning, now with a number.

Two things soften it without excusing it: the flow reference is itself
underpowered at nine runs (±5% interval), and the estimator overstates magnitudes
by a fairly consistent factor (37.8 vs 19.7 here, 26.8 vs 9.4 on `split`), which
looks more like a missing calibration than a random error. Neither is measured
well enough to act on. `stage4` is the positive control and it agrees, so the
comparison mechanism is sound — it is the estimator's confidence that is not.

Detail is in `test/estimation_ladder/README.md`, not here.

PR #822, opened 2026-08-24
test(estimation_ladder): measure the runtime/accuracy ladder properly, on OpenROAD master

oharboe:
**How close can you get to the clock period the flow would give you, without
running the flow?** Measured against running floorplan through global route,
because that is the work this is trying to avoid.

**Where the flow's time goes** on the macro design: global route **374s**,
global placement **200s**, CTS 39s, macro placement 21s -- **668s** total.
Over half is one stage, and the estimator does not skip it so much as run a
version costing 3.5s instead of 374s.

**What that buys**, end to end -- loading the design, running the stages, and
the timing query:

| study | runtime | vs flow | error |
|---|---|---|---|
| `multiplier` (41s flow) | 5.4s | 8x | 1.9% |
| `multiplier_top` (668s flow) | 35s | 19x | 7.9% |
| `multiplier_top`, macro paths only | 20s | 33x | 9.0% |

Overhead -- load plus timing query -- is 2.4s to 4.3s on every rung whatever
it runs, so on the 41s design the whole ladder lives between 4x and 17x. A
cheap flow leaves little to save.

**It answers one question and not the other.** For *what period will this
close at*, yes. For *which path do I fix*, no.

**Macro paths are a separate population and the estimator fails on them.**
They are ~17% faster than the top-level logic, none is among the ten worst
paths, and a slack-ranked sample reached two of them in ninety-nine -- the
design was being graded on its glue logic. Sampled explicitly, the estimator
is 2.4x less accurate on them and its rank correlation goes *negative*: it
orders them backwards. The healthy-looking aggregate was the non-macro
majority outvoting them.

That is not a tuning problem. Across **256 distinct configurations** --
including a whole sweep aimed at the macro objective, which explored 132 the
first sweep never saw -- nothing ranks macro paths better than tau 0.65
against 0.78 for the rest. Macro error also plateaus at ~9% almost
immediately: four times the runtime moves it 0.0904 to 0.0875.

**The estimator runs optimistic** because it places cells and never routes
them, and by nearly a constant amount, so one correction term removes most of
the error. Of ten model families, the best is a constant *added*, not
multiplied. Richer fits all do better on their own design and worse on a new
one -- a Gaussian process reaches zero error on its own design and then does
worse than no correction at all.

**Per-path features can fix ranking, but only within a design.** A random
forest on path features reaches held-out error 0.022 and rank correlation
0.88, against 0.115 and 0.73 for the best estimate-only correction -- and
then transfers to another design worse than doing nothing. The recipe that
[...]

PR #866, opened 2026-08-30
test(estimation_ladder): where is the noise born? Flow-side per-stage variance decomposition and the KPI menu

oharboe:
## Where in the flow is the noise born — and what KPI could a per-PR verdict afford?

`seed_sensitivity.py` measures **sigma_E** — the estimator's per-stage stability — and its docstring is explicit that the flow-side dispersion **sigma_T** is deliberately out of scope. This PR adds that flow-side arm, as a campaign in `test/estimation_ladder/`: hold a stage's input fixed, re-run the *production* stages from there under an admissible perturbation, and read where the flow's variance is actually manufactured. The output is deliberately a **menu, not a verdict**: sigma and the resolvable effect per KPI *candidate*, so choosing what to gate PRs on happens after the compromises are numbers.

### Method

`stage_variance.tcl` chains the production ORFS stage scripts floorplan→grt in **one OpenROAD process** — the `KEEP_VARS=1` sequential sourcing that ORFS's own `flow/scripts/floorplan_to_place.tcl` demonstrates; `load_design` no-ops on an in-memory design — and uses `//fork` (docs/fork.md, #860's machinery) to hang an ensemble off each stage boundary. The spine is paid once; a member costs only its tail:

| arm | forked after | perturbation | members |
|---|---|---|---|
| place | pdn | `GPL_RANDOM_SEED` 2–9 (seed **1 is the default draw** — measured bit-identical to the spine, so it would be a second null) | 8 + null |
| cts | detailed place | clock-period nudge ±1..±4 ps (CTS exposes no seed; the nudge cancels in `min_period = clk_period − slack`, so whatever survives is tool noise — Jeong & Kahng's probe, as in `seed_sensitivity`), read back, error if it did not land | 8 + null |
| grt | cts | `GRT_SEED` 1–8 via `set_global_routing_random` | 8 + null |
| all | pdn | all three at once — the directly measured sigma_total the per-arm decomposition must predict | 8 |

Every member runs the production tail **to grt** and is measured there by `extract_lib.tcl` (the sampling factored out of `extract.tcl`, so the ensembles and the ground truth share one instrument). A stage's spread therefore *includes* whatever the downstream stages amplify it into. Each null member must reproduce the spine bit-for-bit — the harness's free determinism self-test. Members are single-threaded (threads do not survive fork); the driver bounds concurrency with `ORFS_FORK_JOBS`.

### Results

**multiplier_top** (macro array, 36 leaves, ~3.5 h wall at `--jobs 14`). Sigma per KPI candidate, % of the spine's value:

| KPI | place | cts | grt | all |
|---|---|---|---|---|
| achieved (max) | 0.68 | 0.36 | 0.13 | 1.44 |
| top10_mean | 0.72 | 0.37 | 0.11 | 1.44 |
| p95 | 0.74 | 0.38 | 0.14 | 1.48 |
| mean (worst-25% sample) | 0.85 | 0.18 | 0.15 | 1.45 |
| std-cell area | 0.44 | 0.07 | 0.04 | 0.45 |

- **The noise is born early and decays back-to-front**: place ≫ cts ≫ grt. grt from a fixed cts input moves the achieved period by ~0.13% — global route is close to a deterministic transfer function on this design.
- **The validity check earns its keep**: summing the per-arm variances predicts sigma_total = 8.5 ps where the all-levers arm measures 15.6 ps. The bootstrap CI on the interaction term still contains zero at k=8 — *consistent, but the point estimate says the per-stage view under-counts* (the levers compound through timing-driven decisions), which is exactly why the direct total is measured rather than assumed.
- **Area moves too** (±0.44% from placement alone): the repair stages trade area for timing, so a period-only KPI can call a change "improved" that bought speed with area. The KPI is PPA-shaped; **power is recorded equal to std-cell area and left as a `TODO(power)`** until a credible activity assumption exists at grt.
- All determinism nulls reproduced the spine exactly; every clock nudge landed.

**multiplier** (simple, no macros): the same walk finds place at ~0.4%, and cts/grt at the float32 resolution floor (~1e-6%) — the flow below placement is effectively deterministic there. Same guards, all clean.

**The runtime-for-resolution menu** (multiplier_top, `delta_min = 1.96·sigma·√(2/k)` at k=10, KPI = achieved):

| generator | member tail | CPU @ k=10 | resolves |
|---|---|---|---|
| grt-from-cts | ~460 s | ~77 min | 0.12% |
| cts-from-place | ~530 s | ~88 min | 0.31% |
| place-from-floorplan | ~830 s | ~139 min | 0.60% |
| all levers | ~850 s | ~141 min | 1.26% |

Reading it: an ensemble rooted at a late stage is cheap and sharp but sees only that stage's noise; the honest gate variance is the all-levers row. `stage_variance_<design>.json` keeps every member's raw sampled min_periods, so future KPI candidates can be evaluated without re-running the flow.

### Also in this PR

- **`fix(run_executable)`**: `environment_string` embedded `KEY="VALUE"` in shell wrappers, so a make-style reference in an argument value (`PDN_TCL=$(PLATFORM_DIR)/…`) was eaten by shell command substitution before make ever saw it — latent until a script driven through `orfs_run_executable` actually ran the pdn step. Values are now single-quoted.
- `extract.tcl` refactored around the new `extract_lib.tcl`; its JSON gains `clock_period` and area fields (backward-compatible).
- `stage_variance_test.py` holds the aggregation math still in CI on synthetic ensembles (the decomposition, the interaction flag, the k-sizing arithmetic, both guards); the walk itself stays manual.
- README section rendered by `update_readme.py`, present only when the campaign JSONs exist.

### Reproducing

```sh
bazel run //test/estimation_ladder:stage_variance_small  # ~15 min
bazel run //test/estimation_ladder:stage_variance_top    # hours; --jobs per your cores
```

Refs: #860 (fork machinery, the gate campaign), #822 (runtime measurements); Kahng & Mantik (ISQED 2002), Jeong & Kahng, Chan, Kahng & Woo (SLIP 2020); ORFS `GPL_RANDOM_SEED` / `GRT_SEED` ("Useful for perturbation studies").


oharboe:
Superseded by #867, which stacks on this branch and carries all of its commits (the run_executable quoting fix, extract_lib refactor, the variance decomposition campaigns and results) plus the RTL-MP scoring audit. Consolidating review into one PR.

PR #868, opened 2026-08-31
feat(estimation_ladder): measured macro-placement selection on asap7 swerv_wrapper

oharboe:
Built on #867 (merged; this branch is rebased onto main). The audit there ended with a verdict — RTL-MP's objective picks the better of two macro placements at roughly coin-flip probability — and a proposal (`ideas/rtl-mpl.md`): don't trust the objective, replace the *selection*. Fixing the in-anneal estimator is a fool's errand for structural reasons (coarse clustered model mid-anneal, sequence-pair search space, per-run normalization), so the settled architecture is: **RTL-MP is a distribution generator, and every candidate runs through the fast-GPL approximator**. This PR is the derivation work for that selector on the campaign's real target, asap7 `swerv_wrapper` (~28 SRAM macros in three fakeram sizes).

## What it builds

`//test/estimation_ladder:swerv_macro_select` — an `orfs_run()` off the pinned ORFS tree's `swerv_wrapper_synth` that:

1. runs one production floorplan spine,
2. forks k=24 candidates, each running the production `macro_place.tcl` with `-random_seed` appended to ORFS's own constructed argument list,
3. scores each candidate with the estimation ladder's fast non-timing-driven global placement rung — a measurement taken on the far side of the fog that implicitly prices density and congestion,
4. joins, selects on the sampled-path KPI, and materializes the winner as the declared `macro.tcl` output.

The consumer is this package's `swerv_wrapper` flow: the production configuration parsed from the same config.mk (`@orfs_designs`), sharing the baseline's synthesis action via cross-package `previous_stage` (new), plus exactly one change — `MACRO_PLACEMENT_TCL` pointing at `macro.tcl`. A/B is against the untouched `@orfs//flow/designs/asap7/swerv_wrapper` base flow.

## Toolchain: four carried OpenROAD patches

Bumped to OpenROAD master and carried on the `archive_override` while upstream mpl churns. The upstream PRs were parked in favor of issues with minimal reproducers verified against stock master (drop-at-bump notes in MODULE.bazel):

- `0040` re-run segfault fix → The-OpenROAD-Project/OpenROAD#11277.
- `0041` `rtl_macro_placer -random_seed` — deterministic candidate generation (same seed, same placement); a feature, not a bug: returns as a fresh upstream PR when the churn settles.
[...]

PR #881, opened 2026-09-02
feat: the <flow>_estimate report — fast PPA estimate for differential use

oharboe:
Every `orfs_flow()` that runs synthesis grows two targets:

- **`<name>[_variant]_estimate`** — a deterministic, one-to-two-minute estimate of what the design can achieve at its configured floorplan parameters, built **from the synthesis output**: floorplan initialized in-script from the same variables the production stage reads (both init modes), macros placed by the production placer when present, then a non-timing-driven global placement early-stopped at overflow 0.6.
- **`<name>[_variant]_estimate_run`** — the same script as a `bazelisk run` executable with run-time `KEY=VALUE` parameter overrides: **one point in floorplan-parameter space per invocation** (smoke: `CORE_UTILIZATION=10 PLACE_DENSITY=0.6` — floorplan rebuilt at the new point, echo reflects the override).

```sh
bazelisk run //test/smoketest:lb_32x128_asap7_estimate_run -- \
  CORE_UTILIZATION=10 PLACE_DENSITY=0.6 OUTPUT=$PWD/point.json LOG_DIR=$PWD/logs
```

**Division of labor**: this is the *oracle* half of floorplan-parameter automation. bazel-orfs computes points and recommendations — the JSON's `params` object carries the parameters in force under their **exact ORFS variable names**, so a chosen point is directly consumable by ORFS's config.mk pin machinery (`pinAutoFloorplan.py`, OpenROAD-flow-scripts#4487). bazel-orfs never writes config.mk. Constraints imported from that PR's measured negative results: utilization is constraint satisfaction, never blended into a period objective (ρ = −1.0 on gcd for the blend), and "did not resolve" keeps the incumbent.

**Properties**: parallel to the entire physical flow by DAG construction; cannot perturb flow artifacts; deterministic (an estimate delta between commits has no sampling noise of its own); content-addressed (the merge-base baseline for a PR diff is a cache hit; nightly builds are a trend line at ~1% of flow cost); pre-route optimistic **by design, for differential use** where the uniform bias cancels.

**Macro designs**: the JSON carries the macro-path channel (`macro_paths_mean/worst/sampled` — the KPI macro placement actually controls) and `macros_pinned`: unpinned estimate deltas on macro designs carry placement-draw variance; with `MACRO_PLACEMENT_TCL` set (stock ORFS) the draw is held constant and deltas are clean. Racing the pin is #868's business.

**Cadence — estimate always, audit as you can afford, pin occasionally**: the estimate is *purposely* sweep-free (it has no sampling noise to sweep away; the chaos lives in the macro draw and the stages it doesn't run). Pin machinery stores the estimate JSON at pin time — a **receipt** — and the always-on estimate's drift against the receipt is the measured re-pin signal (deltas reported, policy the consumer's, zero new code). And the honest audit economics: an explicit k-seed sweep of a large design is an adorable aspiration — the nightly build over evolving commits *is* the seed sweep, smeared across time; each point confounds improvement with a fresh draw, the trend averages over draws, large effects stand out regardless, and detrended residuals estimate the draw-noise floor for free. When audit compute is starved, the deterministic estimate delta is the only noise-free per-commit measurement in the system.

**`docs/estimate.md` is the thesis**: a flow result is a draw (Kahng & Mantik, ISQED 2002); honest comparison needs seed sweeps; the guided-random-walk model shows a fast imperfect gate plus an overnight seed-sweep audit reaches the PPA goal with orders of magnitude less gate compute; calibration provenance (measured vs pending — E14 backtest named); own-your-scoring-function guidance; stated limits (a speedometer, not a map).

This PR consolidates the *lessons* of the exploratory campaign work into one thesis; the doc's "Why the estimator is shaped this way" section gives the design rationale without the number-dump, and its "Provenance, trust, and breadcrumbs" section says plainly what a thesis owes and doesn't: the numbers were observed once on one machine with evolving instruments — **do not trust them, or the archived data, without re-measuring** — with pointers to where the evidence and its flaws live, so a proper follow-up study can be done right.

Tests: `lb_32x128_estimate_test` asserts the JSON contract; smokes on the smoketest designs. Pending before ready: capstone period-fidelity numbers into the calibration section, one macro-design smoke on the freed campaign machine.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -13,6 +13,7 @@ exports_files([
     "compute_slack_margin.tcl",
     "config_mk_parser.py",
     "deploy.tpl",
+    "estimate.tcl",
     "html_timing_report.tcl",
     "make.tpl",
     "mock_area.tcl",
--- a/README.md
+++ b/README.md
@@ -42,6 +42,7 @@ and easily actionable github issues for the OpenROAD and ORFS maintainers.
 | Upgrade bazel-orfs or ORFS | [Upgrade bazel-orfs](#upgrade-bazel-orfs) |
 | Override configuration variables | [Override configuration variables](#override-configuration-variables) |
 | Diagnose a build failure on my host | [docs/debugging.md](docs/debugging.md#host-platform--older-distributions) |
+| Fast PPA estimate / gate a PR against merge-base | [docs/estimate.md](docs/estimate.md) |
 
 ## Requirements
 
--- a/docs/estimate.md
+++ b/docs/estimate.md
@@ -364,6 +364,25 @@ ranking that candidate on wirelength would have shipped a design that
 does not build. Feasibility is a separate outcome from score, and
 "fraction of candidates infeasible" is worth tracking as its own number.
 
+**A near miss at small n is often not near anything.** The best of the
+cheap readouts landed at rho +0.39 [-0.01, +0.72] -- short of clearing
+zero by 0.01, which invites one more turn of the crank. Taking that turn
+(doubling the readout's support) moved it to +0.42 [-0.02, +0.73] at
+2.3x the cost, and it *still* straddled zero. At n=23 those two are the
+same measurement, and the 0.01 shortfall was where the interval happened
+to fall rather than a gap that tuning could close. Before spending on a
+variant of a near miss, ask what the interval width says the measurement
+can resolve; here it could not resolve the difference the variant was
+built to produce.
+
+**A coarsening rung pays for every exception it makes.** The same
+experiment held instances near macro pins out of the clusters to keep
+their pin geometry exact. Going from one hop to two raised that
+exempt set from 1458 instances to 2380 and cut the speedup over the flat
+scorer from 4.3x to 1.8x -- most of the rung's reason to exist, spent on
+fidelity that bought no ranking. Exemptions from an abstraction are
+charged at the un-abstracted rate, so they need their own budget.
+
 **A confidence interval spanning [-1, +1] is not a pass.** On a
 four-candidate population a gate phrased as "rho lands inside the
 reference interval" reported PASS at rho +0.80 with a bootstrap interval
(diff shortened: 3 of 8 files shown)


oharboe:
Pushed `d378625` — the last E12 loose end, and it's a negative that changes this PR's own recommendation.

The write-up closed by naming the macro-cone readout as the one live thread: **+0.39 [-0.01, +0.72]**, best of the HPWL family, short of clearing zero by 0.01 — and recommended wider hop counts as the cheap next step. That step has now been run.

| | 1 hop | 2 hops |
|---|---|---|
| cone rho vs `macro_paths_mean` | +0.39 [-0.01, +0.72] | +0.42 [-0.02, +0.73] |
| full clustered HPWL rho | +0.26 [-0.22, +0.66] | +0.32 [-0.13, +0.68] |
| agreement with the flat rung | +0.66 [+0.34, +0.86] | +0.71 [+0.41, +0.89] |
| instances held out of clusters | 1458 | 2380 |
| `gpl` time | 14.7 s | 34.4 s |
| cheaper than flat | 4.3x | 1.8x |

+0.02 for 2.3x the cost, interval still straddling zero. At n=23 those are the same measurement, so the 0.01 shortfall was where the interval happened to fall — not a gap tuning could close. Recommendation 1 now reads "do not chase the cone" instead of "chase the cone".

The cost half is a separate finding worth keeping: holding instances near macro pins out of the clusters is what preserves their exact pin geometry, and the second hop raises that exempt set from 1458 to 2380, cutting the speedup from 4.3x to 1.8x. **A coarsening rung is charged the un-abstracted rate for every exception it makes** — most of this rung's reason to exist, spent on fidelity that bought no ranking.

Two entries added under "Negative findings of importance"; table and recommendation updated in `ideas/e12-clustered-scorer.md`. All 16 quoted figures are asserted against `e12_verdict_swerv_remeasured.json` and the 2-hop `verdict.json` rather than transcribed by hand.

With this, E12 is closed out and everything from ORFS#4492 has landed here.

PR #888, opened 2026-09-02
fix(env): prefix WORK_HOME with external/<repo> for external designs

oharboe:
Found by running an ORFS design interactively from a bazel-orfs workspace:

```
$ bazelisk run @orfs//flow/designs/asap7/gcd:gcd_synth -- open_synth
...
make: *** No rule to make target 'open_1_synth.odb', needed by 'open_synth'.  Stop.
```

`required_arguments()` built `WORK_HOME` as `"./" + ctx.label.package`, correct only for a design in the main repository. An external repository's files live under `external/<repo>/` in bazel's exec root, in its runfiles layout, and in the tree `deploy.tpl` lays out for `bazelisk run` — so `@orfs//flow/designs/asap7/gcd` got `./flow/designs/asap7/gcd`, a path holding none of that design's results.

## Why the error names a prerequisite, not a directory

ORFS declares its interactive targets from a wildcard over the results directory (`scripts/variables.mk`):

```make
RESULTS_ODB = $(notdir $(sort $(wildcard $(RESULTS_DIR)/*.odb)))
```

and then `$(foreach file,$(RESULTS_ODB),$(eval $(call OPEN_GUI,$(file),ODB_FILE)))`. With `WORK_HOME` wrong the wildcard matches nothing at parse time, so `open_1_synth.odb` is never declared — while `open_synth` still exists, because `OPEN_GUI_SHORTCUT` is unconditional. Hence a missing *prerequisite* for a target that does exist, which reads like a bug in the shortcut rather than a wrong path.

`gui_synth` and `web_synth` fail the same way, for the same reason.

## The fix

`orfs_test` and `orfs_run_executable` had already hit this and each carried its own copy of the same six lines. All three now share one helper, expressed in terms of the `_workspace_prefix()` already in `environment.bzl`, so they cannot drift apart again — and `_work_home()` (the build-action variant) is folded onto it too.

The main-repository case is unchanged by construction: the prefix is empty there, leaving `"./" + package`.

## Verified

- `//test:work_home_test` — 4 tests; reverting either half of the fix fails 3 of them
- `//test:keep_modules_test` and the rest of the suite still pass
- `bazelisk build --nobuild --keep_going //...` — 1,004 targets analyzed, 0 errors
- `//:fix_lint` clean

## Note on scope

This is a prerequisite for building `@orfs` designs from here, but it's an independent bug on `main` and reviewable on its own — no dependency on #887.

### DIFF
--- a/private/environment.bzl
+++ b/private/environment.bzl
@@ -23,16 +23,36 @@ def sdc_arguments(ctx, short = False):
         return {"SDC_FILE": file_path(sdc, short)}
     return {}
 
+def work_home_relative(ctx):
+    """WORK_HOME relative to the root of a runfiles or deployed tree.
+
+    An external repository's files live under `external/<repo>/` in
+    bazel's exec root, in its runfiles layout, and in the tree deploy.tpl
+    lays out for `bazelisk run`. So a target in, say,
+    @orfs//flow/designs/asap7/gcd has to look for its results under
+    external/orfs+/flow/designs/asap7/gcd -- under the main repo's own
+    package path there is nothing of that design at all.
+
+    Args:
+      ctx: The rule context.
+
+    Returns:
+      A path string relative to the tree root, with no leading "./".
+    """
+    return "/".join([
+        part
+        for part in [_workspace_prefix(ctx), ctx.label.package]
+        if part
+    ])
+
 def _work_home(ctx):
     # For external repo targets, declared files are placed under
     # bin/external/<repo>/<package>, but genfiles_dir.path is just bin/.
     # Use the declared file path of a known output to derive the correct prefix.
     parts = [ctx.genfiles_dir.path]
-    if ctx.label.workspace_name:
-        parts.append("external")
-        parts.append(ctx.label.workspace_name)
-    if ctx.label.package:
-        parts.append(ctx.label.package)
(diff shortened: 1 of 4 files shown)


oharboe:
Correction on ordering, from trying to verify this end-to-end.

I claimed above that #888 is reviewable independently of #887. That is true of the *code* — this bug predates #887 and the fix stands alone — but it is **not** true of the verification. On `main`, `@orfs` designs do not load at all:

```
$ bazelisk run @orfs//flow/designs/asap7/gcd:gcd_synth -- -n open_synth
ERROR: .../external/orfs+/flow/designs/asap7/gcd/BUILD, line 11, in <toplevel>
    design(
  .../flow/designs/design.bzl, line 74, in design
    orfs_design(
Error: orfs_design() got unexpected keyword argument: blender
```

That is exactly the drift #887 exists to remove. So the wrong `WORK_HOME` was unreachable from `main` — you cannot get far enough to hit it — which is why it survived unnoticed, and why #887 is what surfaced it.

Practical consequence: **#887 lands first**, then this. I'm running the differential (`-n open_synth` before and after, same output base, one commit apart) on a tree carrying both and will post the result here rather than assert it.

PR #890, opened 2026-09-02
feat(floorplan): derive a design's floorplan, and pin it into config.mk

oharboe:
> **A demonstration, not a core feature.** A worked idea — that the floorplan numbers in a `config.mk` are measurable rather than guessable, and roughly what measuring them costs — kept runnable so the idea is concrete. Anyone who wants it in earnest is expected to reimplement it inside their own project, against their own designs, provisioning and definition of "better". Read the ladders, tie bands and winner rule as one set of choices that worked on a handful of ORFS designs, not as an interface to build on.
>
> What this PR does need to be is **additive and inert**: nothing in the normal flow may depend on it, and both targets are `manual`. That part is measured below.

Rebased onto `main`, verified additive, and the pin now works on a design in another repository.

## The target this enables

```sh
bazelisk run @orfs//flow/designs/asap7/gcd:gcd_auto_floorplan_pin ~/ORFS
```

The config path the target passes is relative to the repository the design lives in, so for an `@orfs` design it is relative to an ORFS checkout — not to the bazel workspace you ran from, and not to the fetched archive, which is read-only. The script resolved it against `BUILD_WORKSPACE_DIRECTORY`, correct only when design and workspace are the same repo, so the command above would have looked under the bazel-orfs workspace and exited. There is now an optional third positional argument for the checkout to edit, which wins over the environment.

## Run for real against a pristine ORFS checkout at the pinned commit

```
derive   315.9 s wall, 14 candidates (6 util + 5 addon + 3 aspect) at AF_JOBS=2, exit 0
pin      exit 0
```

```diff
-export CORE_UTILIZATION         = 65
+export CORE_UTILIZATION         = 78
```

One line, and that is correct rather than partial. The winner — `util 78, aspect 1, margin 0.5, density 0.35, addon 0, drc 0`, achieving 349.7 ps against a 310 ps target — already agrees with `config.mk` on aspect, margin and density, so only utilization differs. Position, spacing and the assignment operator are preserved; no generated block, no markers.

## Additive, measured rather than asserted

```
bazelisk query '@orfs//flow/designs/...:*'    4321 targets, 0 errors
vs main's 4135                                0 lost, +186 added
every one of the 186 is auto-floorplan        62 designs x {auto_floorplan.json, _data, _pin}
bazelisk test //test:all //:all               124/124 pass
```

`design(designs)` is now **optional** (`= None`, with `_auto_floorplan()` returning early). As written it was positionally required, which would break any caller on the older two-argument signature.

## What the rebase surfaced

**Patch `0047` would have broken at the ORFS transition.** It edits ORFS's own `flow/BUILD` to export `scripts/flow.tcl`, so it stops applying the moment ORFS deletes that file — which is exactly what The-OpenROAD-Project/OpenROAD-flow-scripts#4501 does, and the same trap patch `0046` hit. `_GENERATE_FLOW_BUILD` now exports `scripts/flow.tcl` too, and `//test:orfs_flow_build_test` requires it. Both mechanisms are needed: the patch covers an ORFS that still ships `flow/BUILD`, the generated file covers one that does not.

**`MODULE.bazel` conflicted, as expected.** `main` moved ORFS from a `bazel_dep` to an extension-created `http_archive`, so the `archive_override` this branch added `0047` to no longer exists; the patch moved to `ORFS_PATCHES` in `orfs_source.bzl`.

**Three files were not `black`-clean**, so CI's Lint step would have failed as-is — 9 lines of reformatting in `check_pareto.py`, `check_pareto_test.py` and `auto_floorplan_flow_test.py`, none of it mine.

## Scope worth knowing: 62 designs, not all of them

`design()` looks up `DESIGNS[<last two path components>]`, so ORFS's three nested block designs — `asap7/riscv32i-mock-sram/fakeram7_256x32`, `gf180/uart-blocks/uart_rx`, `ihp-sg13g2/i2c-gpio-expander/I2cDeviceCtrl` — yield a key that does not exist and get no targets, while their **parents** do. That is right: a block's floorplan is derived as part of its parent's flow. Pinned by a test so the silence is a decision rather than a surprise, and so widening the lookup later is deliberate.

## `rules-base.json` comes from CI, not from a local run

New section in `docs/auto_floorplan.md`, because pinning a floorplan changes QoR and the bounds stop describing the design. Regenerating them locally bakes in local noise: yosys is not idempotent, ORFS upstream treats that drift as expected, and tool versions and thread counts move the numbers besides.

The supported path is ORFS's `set-new-golden` `repository_dispatch`, which runs `flow/util/updateRules.py --commitSHA <sha>` against the metrics **CI itself measured** for that commit and opens a draft PR; `genRuleFile.py` / `genAllRuleFiles.sh` are what that path uses underneath. `mock-alu` is the concrete case — one of 27 bounds missed by 0.9% purely from an OpenROAD bump.

So after pinning: land the `config.mk` change, let CI measure it, request the golden update. Read a local metrics diff for direction and magnitude, never as values to commit.

## Not verified

No derivation on a large design. `AF_JOBS` is memory-bound — the branch already records an 18 GB-per-candidate OOM after four hours on `swerv_wrapper` — and provisioning is a judgement call for whoever runs it. The 315.9 s above is `gcd`, the smallest design in the flow.

And CI does not exercise the derivation at all: both targets are `manual`, so `bazelisk test ...` skips them. Green here means "nothing existing broke and the unit tests pass" — which is precisely the additivity claim, and the right thing for a demonstration to be held to. The end-to-end evidence is the local `gcd` run above, not CI.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -16,8 +16,12 @@ exports_files([
     "orfs_design_builds.bzl",
     "record_orfs_builds.py",
     "run_executable.py",
+    "auto_floorplan.tcl",
+    "auto_floorplan_candidate.tcl",
+    "auto_floorplan_flow.tcl",
     "bump.py",
     "bump_impl.py",
+    "check_pareto.py",
     "compute_floorplan_shape.tcl",
     "compute_slack_margin.tcl",
     "config_mk_parser.py",
@@ -30,6 +34,7 @@ exports_files([
     "open_html.sh",
     "package_stage.py",
     "parallel_synth.mk",
+    "pin_auto_floorplan.py",
     "power.bzl",
     "power.tcl",
     "power_base.tcl",
@@ -268,4 +273,47 @@ compile_pip_requirements(
     name = "requirements_dev",
     src = "requirements_dev.in",
     requirements_txt = "requirements_dev_lock_3_13.txt",
+)
+
+# --- floorplan derivation ------------------------------------------------
+#
+# compute_floorplan_shape.tcl predicts CORE_UTILIZATION and CORE_MARGIN
+# from synth-stage area. This derives them instead, by running the
+# production flow per candidate floorplan and reading what it achieved.
+# The two are not rivals yet: the derivation is what can grade the
+# heuristic, per design, because it produces the flow's own answer on the
+# same ladder.
+
+py_binary(
+    name = "pin_auto_floorplan",
+    srcs = ["pin_auto_floorplan.py"],
+    visibility = ["//visibility:public"],
+)
+
+py_test(
+    name = "pin_auto_floorplan_test",
+    srcs = ["pin_auto_floorplan_test.py"],
+    deps = [":pin_auto_floorplan"],
+)
+
+py_test(
+    name = "check_pareto_test",
+    srcs = ["check_pareto_test.py"],
+    data = ["check_pareto.py"],
+)
+
+# Fails if auto_floorplan_flow.tcl stops being flow.tcl's tail. flow.tcl
+# is read from the pinned @orfs archive so the comparison is against the
+# sequence that actually runs; see patches/0047.
+py_test(
+    name = "auto_floorplan_flow_test",
+    srcs = ["auto_floorplan_flow_test.py"],
+    data = [
+        "auto_floorplan_flow.tcl",
+        "@orfs//flow:scripts/flow.tcl",
+    ],
+    env = {
+        "AF_FLOW_TCL": "$(location auto_floorplan_flow.tcl)",
+        "ORFS_FLOW_TCL": "$(location @orfs//flow:scripts/flow.tcl)",
+    },
 )
--- a/auto_floorplan.tcl
+++ b/auto_floorplan.tcl
@@ -1,0 +1,658 @@
+# Floorplan derivation: measure the floorplan shape instead of guessing it.
+#
+# Run by <name>_auto_floorplan_data, off a design's synthesised netlist.
+# Races candidate outlines and densities, each evaluated by running the
+# production flow to finish, and writes the winner and every candidate to
+# $AF_EVIDENCE. <name>_auto_floorplan_pin turns that into config.mk
+# values. Nothing here runs during an ordinary build.
+#
+# WHAT IS BEING RETIRED
+#
+# CORE_UTILIZATION (or DIE_AREA/CORE_AREA), CORE_ASPECT_RATIO and
+# PLACE_DENSITY (or PLACE_DENSITY_LB_ADDON) are the last hand-carried
+# numbers in a config.mk. Each is a human prediction of downstream
+# behaviour standing where a measurement should be: the right utilization
+# is "the smallest core in which this netlist still closes", which is a
+# question about repair_design, CTS and routing -- none of which have
+# happened when the number is read.
+#
+# THE ORACLE IS THE FLOW
+#
+# An earlier version scored candidates with a fast pre-route proxy. It
+# does not rank the utilization axis: measured against the finished flow
+# over a utilization ladder, rho = -1.000 on gcd (perfectly inverted) and
+# +0.100 on gcd-ccs -- the two designs whose noise floor is small enough
+# for the question to have an answer. So each candidate runs the real
+# flow and reports what it achieved. That is expensive, and it is why
+# this is a job you run rather than something the flow does.
+#
+# Every ladder is a dimensionless fraction. There is no per-design
+# constant anywhere in this file.
+
+proc af_env { name { default "" } } {
+  if { [info exists ::env($name)] && $::env($name) ne "" } {
+    return $::env($name)
+  }
+  return $default
+}
+
+proc af_log { msg } {
+  puts "AUTO_FLOORPLAN: $msg"
+}
+
+# Fractions of the incumbent utilization. Above 1.0 is a smaller core;
+# one point below is kept so the walk can report that loosening was
+# better rather than only ever being able to tighten.
+set ::af_util_ladder [af_env AF_UTIL_LADDER {0.9 1.0 1.1 1.2 1.3 1.4}]
+
+# Fractions of the headroom above the measured density lower bound. This
+# is exactly what PLACE_DENSITY_LB_ADDON means; the ladder makes it
+# measured instead of guessed.
+set ::af_addon_ladder [af_env AF_ADDON_LADDER {0.00 0.05 0.10 0.15 0.20}]
+
+# Aspect ratios to try at the winning area point. Often a tie, which is a
+# result: the folklore value gets graded rather than assumed.
+set ::af_aspect_ladder [af_env AF_ASPECT_LADDER {0.8 1.0 1.25}]
+
+# Placer seeds for the noise floor. Eight rather than three: the range of
+# a small sample is a poor estimator of spread, biased low and with an
+# expectation that grows with n, so a range-based floor would move every
+# threshold purely by changing the seed count. These run concurrently.
+set ::af_tie_seeds [af_env AF_TIE_SEEDS {1 2 3 4 5 6 7 8}]
+
+# The exchange rate between area and period: one percent of achieved
+# period is worth this many percent of core area. A product decision --
+# how much silicon a percent of speed is worth -- not a tuning constant,
+# which is why it is a rate and not a threshold.
+#
+# Validated once against an asap7 sweep rather than searched per design:
+# lambda = 3 rejects the trades that gave up a lot of period for little
+# area (ethmac +13.4% period for -16.7% area, aes +5.6% for -8.7%) and
+# accepts every win and every cheap trade (gcd-ccs +0.78% for -24.2%,
+# mock-alu +0.92% for -28.8%, aes_lvt +1.37% for -28.5%). Every verdict
+# is unchanged for lambda in [2, 5], so this is the shape of a policy
+# rather than a fitted number.
+set ::af_lambda [af_env AF_LAMBDA 3.0]
+
+# Width of the handover around the target period, as a fraction of the
+# clock. Verdicts unchanged for tau in [0.01, 0.10].
+set ::af_tau_frac [af_env AF_TAU_FRAC 0.02]
+
+# ---------------------------------------------------------------------
+# The objective
+# ---------------------------------------------------------------------
+
+# ln(1 + e^x), computed so neither tail overflows.
+proc af_softplus { x } {
+  if { $x > 0 } {
+    return [expr { $x + log(1.0 + exp(-$x)) }]
+  }
+  return [expr { log(1.0 + exp($x)) }]
+}
+
+# What the derivation minimises:
+#
+#   P_eff = T + tau * ln(1 + e^((p - T)/tau))
+#   J     = ln(area) + lambda * ln(P_eff)
+#
+# P_eff is a smooth one-sided penalty: it tracks the achieved period p
+# when the design is slower than its target T and flattens to T when it
+# is faster. Speed beyond what the SDC asked for has no value, so area
+# spent buying it is never repaid; a design short of target pays the full
+# exchange rate for every percent it gives up.
+#
+# There is no regime switch to get wrong. The same expression minimises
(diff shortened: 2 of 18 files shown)


oharboe:
## Campaign status and hand-off

Everything measured so far, so this PR is self-contained and someone
(human or another session) can pick it up cold. Overnight results will be
added as a follow-up comment.

All runs on a 32-core / 122 GB host. asap7 only.

### Measured results

**`aes_lvt`** — the design that was going to be the single-concern
existence proof. Both arms measured by the production flow, same commit,
same day:

| | baseline | derived | |
|---|---:|---:|---|
| `finish__timing__setup__ws` | +9.00 | −0.46 | closure lost |
| achieved period (clock − WS) | 351.0 | 360.46 | +2.7% |
| `finish__design__core__area` | 3480.4 | 2484.2 | −28.6% |
| `finish__design__instance__area` | 1308.8 | 1300.4 | −0.6% |
| `detailedroute__route__wirelength` | 51683 | 49648 | −3.9% |
| `finish__power__total` | 0.0875 | — | |
| `detailedroute__route__drc_errors` | 0 | 0 | |

Derived winner: `CORE_UTILIZATION 40 → 56`, `PLACE_DENSITY 0.65 → 0.611`.
Not filed. A 28.6% smaller core bought by spending all 9 ps of setup
margin and crossing zero is not an existence proof of anything good.

**`cva6`** — clean negative, and the useful kind:

```
utilization phase: 70 minimises the objective
  (2.338e+04 um2, achieved 989.4; incumbent 2.338e+04 um2, achieved 989.4)
objective: winner J 30.81049, incumbent J 30.81049, noise 0.01170
the winner does not beat the incumbent by more than the noise floor
  could account for; keeping the incumbent
```

Utilizations 63–98 raced; the design's existing 70 won. Nothing pinned,
nothing can regress. Eliminations: top rung on `FLW-0024`, two density
candidates on `PSM-0069`. Winner record: util 70, aspect 1, margin 2,
density 0.875, core 23378.7 um2, achieved 989.4 against a 1000 ps target,
DRC 0.

Caveat: this run predates the 0a fix, so its "incumbent" was
util-70-at-addon-0.10 rather than the design's stated `PLACE_DENSITY`.
Because the verdict is *change nothing*, 0a cannot have caused a
regression here — the worst case is a missed opportunity. It is **not** a
validated statement about `cva6` as configured.

**Pre-route proxy vs the finished flow** (why the oracle is the full flow):

| design | ρ (proxy vs flow) | noise floor |
|---|---:|---:|
| gcd | −1.000 | 1.1% of clock |
| gcd-ccs | +0.100 | 0.3% of clock |

### Measured noise floors (`delta_tie`, 2σ over 8 placer seeds)

| design | 2σ | as % of clock |
|---|---:|---:|
| aes_lvt (run 1) | 4.15 ps | 1.15% |
| aes_lvt (run 2) | 6.87 ps | 1.91% |
| cva6 | 9.26 ps | 0.93% |

**The floor is itself imprecise.** Two runs of the same design at the same
coordinates gave 4.15 and 6.87 — a 65% swing. Any verdict that turns on a
difference of this size is not resolvable with 8 seeds, which is what
withdrew 0b and what makes 0d's "what margin counts as missing" hard.

### Wall-clock, for planning

| run | time |
|---|---|
| `aes_lvt` derive, `AF_JOBS=2` | 6526 s |
| `aes_lvt` re-derive (0a verification) | 6871 s |
| `cva6` derive, `AF_JOBS=2`, ~20 candidates | ~17 h |
| `swerv_wrapper`, `AF_JOBS=8` | **OOM-killed at 14515 s** |

Roughly 20 candidates per design: 6 utilization + 8 seeds + 5 density + 2
aspect. Phases are serialised, so wall time is
`ceil(phase_size / AF_JOBS)` batches summed over phases — which is why
`AF_JOBS` matters so much more than it looks.

### Defects

| | | status |
|---|---|---|
| 0a | incumbent raced at a density the design never states | fixed, verified |
| 0b | 9 ps oracle-vs-flow gap | withdrawn — inside the noise |
| 0c | checked-in `rules-base.json` lags the toolchain | not a bug; constrains method |
| 0d | the objective accepts a lost timing closure | **open — blocks this PR** |

Full write-up in `docs/plans/auto_floorplan_ideas.md` §0.

0c, briefly, because it affects anyone comparing QoR in ORFS: asap7
`rules-base.json` files were last regenerated between one and seven weeks
ago. On `aes_lvt` the checked-in bound implies `ws = 0.0` where a fresh
run on the same commit measures `+9.00`. Nothing is malfunctioning — the
guard is one-sided and a design getting faster is not a regression — but a
`rules-base.json` diff cannot serve as evidence for a change without
separating that drift, and back-solving a baseline out of a padded bound
will invert verdicts. It did here: it turned a lost closure into an
apparent 2.5% speedup.

### What is verified, and what is not

Verified in this PR:
- all unit tests pass, including the `flow.tcl` drift guard against the
  pinned `@orfs` archive
- `patches/0047` applies cleanly
- `@orfs//flow/designs/asap7/gcd:gcd_auto_floorplan_{data,pin}` analyze,
  so the targets do materialise for designs that ship no wiring
- bazel-orfs's own DSL tests still pass after `design()` gained `designs`

**Not verified: no candidate has executed through the ported plumbing.**
The scripts previously found each other through ORFS's `SCRIPTS_DIR` and
now go through `AF_CANDIDATE_TCL` / `AF_FLOW_TCL`. If that wiring is
wrong, everything above still passes and the first real derivation fails.
This is the largest known risk in the PR.

### Next steps, in order

1. Smoke-derive one small design (`gcd`) from bazel-orfs, to exercise the
   ported plumbing end to end.
2. Settle 0d. Recommendation: closure as a survival filter alongside DRC
   rather than a term in `J` — but it needs a stated margin, and the noise
   floors above are why that is not obvious.
3. Reuse `config_mk_parser.py` for the pin's *read* side. Its writer must
   stay bespoke (it preserves `export`, spacing and assignment operator),
   but shipping a second config.mk reader into this repo contradicts the
   argument for owning the DSL here.
4. Grade `compute_floorplan_shape.tcl`: ρ between its prediction and the
   flow's outcome, per design, from ladder data the derivation already
   produces.
5. `aes-mbff` derives nothing — every candidate fails `IFP-0018` "Unable
   to find site: asap7sc7p5t_pg". The platform adds that site when
   `CLUSTER_FLOPS=1`; the staged `1_synth.odb` does not carry it. The
   normal flow works, so it is specific to the derivation.
6. `<name>_pareto_test` still describes the per-build feature that the
   derive/pin split replaced.

### Known gaps in the corpus

- `coralnpu` cannot be derived from bazel at all: it ships `config.mk`,
  `constraint.sdc` and `rules-base.json` but no `BUILD`, so there is no
  target. Unrelated to this work.
- Parallel synthesis is broken for `riscv32i`, `riscv32i-mock-sram` and
  `aes-block` (`per-module checkpoint missing: partition_*_canonical.rtlil`),
  which costs those designs from every table. Before the floorplan stage,
  so unrelated.
- Minor: the evidence file records `tie_n` as null rather than the seed
  count.

oharboe:
## Overnight: `swerv_wrapper`

First derivation to run with the 0a fix, so its incumbent is the
design's actual configuration rather than a substituted density.

Re-run at `AF_JOBS=4` after the `AF_JOBS=8` attempt was OOM-killed
(see the provisioning commit). Exit code `0`, 16.5 h.

```
AUTO_FLOORPLAN: incumbent: utilization 30, aspect 1.0, margin 1.0, addon 0.2
AUTO_FLOORPLAN: running 6 candidates, 4 at a time
AUTO_FLOORPLAN: candidate u0p9 eliminated (exit 139); last lines of its log:
AUTO_FLOORPLAN: candidate u1p1 eliminated: 2 DRC violations
AUTO_FLOORPLAN: candidate u1p2 eliminated: 2 DRC violations
AUTO_FLOORPLAN: running 7 candidates, 4 at a time
AUTO_FLOORPLAN: noise floor over 8 seeds: 2 sigma = 67 (4.19% of the clock period)
AUTO_FLOORPLAN: utilization phase: 39 minimises the objective (1.229e+05 um2, achieved 1556; incumbent 1.598e+05 um2, achieved 1547)
AUTO_FLOORPLAN: running 5 candidates, 4 at a time
AUTO_FLOORPLAN: candidate d0p00 eliminated: 1 DRC violations
AUTO_FLOORPLAN: candidate d0p05 eliminated: 1 DRC violations
AUTO_FLOORPLAN: candidate d0p10 eliminated: 1 DRC violations
AUTO_FLOORPLAN: running 2 candidates, 4 at a time
AUTO_FLOORPLAN: candidate a0p8 eliminated (flow failed: PDN-0179)
AUTO_FLOORPLAN: objective: winner J 33.85952, incumbent J 34.12527, noise could account for 0.04497
AUTO_FLOORPLAN: winner a1p25: utilization 39, aspect 1.25, density 0.394, core 1.23e+05 um2, achieved 1531 against a target of 1600
AUTO_FLOORPLAN: evidence written to bazel-out/k8-fastbuild/bin/flow/designs/asap7/swerv_wrapper/auto_floorplan.json
```

Posted automatically when the run finished; nobody was watching it.

oharboe:
### Reading the `swerv_wrapper` result

The first derivation in this campaign that produces a **genuine Pareto
move**, and the first on a design with macros.

| | incumbent | winner | |
|---|---:|---:|---|
| `CORE_UTILIZATION` | 30 | 39 | |
| `CORE_ASPECT_RATIO` | 1.0 | 1.25 | inside noise, see below |
| `PLACE_DENSITY_LB_ADDON` | 0.2 | 0.2 | unchanged |
| core area (um2) | 159800 | 122977 | **−23.0%** |
| achieved period (ps) | 1547 | 1530.6 | tie — see noise floor |
| target | 1600 | 1600 | both comfortably under |
| DRC | — | 0 | |

`J` 34.12527 → 33.85952, so ΔJ = **−0.266 against a noise band of
0.04497** — about 5.9× the band. Accepted, and not marginally.

**0d did not bite here.** The winner lands at 1530.6 against a 1600 ps
target, so there is no closure to lose and the objective's willingness to
price one away never came into play. That is luck about this design, not
evidence that 0d is benign.

Three things worth reading carefully:

**The area win is the whole result; the aspect change is noise.** Core
area at aspect 1.25 (122977) is within 0.1% of the utilization phase's
aspect-1.0 point (~122900), and the 25 ps period difference is well
inside the 67 ps noise floor. The headline is the utilization move from
30 to 39, alone. The aspect ladder graded folklore and found nothing —
which is still a result, just not this one.

**Density folklore was right.** The winner's addon (0.2) is the
incumbent's. Three of five density candidates were eliminated on DRC and
the survivors did not beat what the design already had.

**The guards are load-bearing on a real design.** Seven of roughly twenty
candidates were eliminated: two utilization candidates on DRC, three
density candidates on DRC, one on `PDN-0179`, and one on exit 139 — a
segfault, which the driver treats as an eliminated candidate rather than
a crash. On the small designs the hard filters looked decorative; here
they removed a third of the search space, and a scorer without them would
have ranked DRC-violating floorplans as winners.

### Noise floors, updated

| design | 2σ | as % of clock |
|---|---:|---:|
| aes_lvt (run 1) | 4.15 ps | 1.15% |
| aes_lvt (run 2) | 6.87 ps | 1.91% |
| cva6 | 9.26 ps | 0.93% |
| **swerv_wrapper** | **67.0 ps** | **4.19%** |

swerv's floor is an order of magnitude above the small designs in
absolute terms and four times higher as a fraction of clock. Any period
conclusion on this design is weak; the area number is exact, since the
same coordinates give the same core every time.

### What this does not yet establish

This is the **derivation's own measurement**, not the flow's verdict.
The `aes_lvt` lesson was precisely that those can differ enough to invert
a conclusion, and the honest check is to apply the pin and measure both
arms with the production flow. That has not been done for
`swerv_wrapper`, so `−23% core area at flat period` is a strong
indication, not a demonstration.

Correction to an earlier framing: this run is not trustworthy *because
of* the 0a fix. `swerv_wrapper` states `PLACE_DENSITY_LB_ADDON`
directly, so its incumbent always took the correct path and 0a never
applied to it. It is the first post-fix run, but the fix is not what
makes it clean.

PR #909, opened 2026-09-04
docs(plans): wire up every asap7 and sky130 ORFS design, starting from coralnpu

oharboe:
## Plan for approval, no code changes

`bazelisk build @orfs//flow/designs/asap7/coralnpu:CoreMiniAxi_route` fails at analysis: `@orfs//flow/designs/src/coralnpu` has no package. ORFS added the design in #4474 without a BUILD (ORFS does not run bazel), the recorded-BUILD set was captured at the same pin and so never held it, and the generator only covers `config.mk` directories. Every new ORFS source directory will fail the same way.

Behind it sits a second failure: coralnpu sets `SYNTH_HIERARCHICAL=1` with no `SYNTH_KEEP_MODULES`, so the parallel-synth path dies with `per-module checkpoint missing` (the case 2bde78e demoted to a warning).

The plan also sweeps every `config.mk` under `flow/designs/asap7`, `sky130hd` and `sky130hs` at the pin: 34 configs, 32 `_route` targets, analysed with `--nobuild --keep_going` in 28 s. Only coralnpu fails analysis; five designs carry the hierarchical-without-list problem; swerv_wrapper's dropped `$(CLKGATE_MAP_FILE)` is benign.

Proposed, one PR per concern:

1. Generate `files("verilog")` / `files("include")` absent-only for `flow/designs/src/**` directories with no BUILD, with the test inversions that implies.
2. A CI step that *analyses* the design flow targets, not just loads their packages.
3. Serial-synth fallback when a hierarchical design has no kept-module list, instead of 32 partitions that cannot succeed.
4. Run the coralnpu route flow; ORFS follow-ups (pin `SYNTH_KEEP_MODULES` for coralnpu) on request.

See `docs/plans/orfs-designs-wired-up.md`.

### DIFF
--- a/docs/plans/orfs-designs-wired-up.md
+++ b/docs/plans/orfs-designs-wired-up.md
@@ -1,0 +1,273 @@
+# Plan: every asap7 and sky130 ORFS design wired up
+
+> Written by Claude with Øyvind Harboe from one failing build and an
+> analysis-only sweep of every design under `flow/designs/asap7`,
+> `flow/designs/sky130hd` and `flow/designs/sky130hs` at the pinned ORFS
+> (`8c0616910`, 2026-09-01). Nothing in this document is implemented yet;
+> section 7 is the proposal for approval. Where a claim rests on a build
+> that has not been run, it says so.
+
+---
+
+## 1. What happened
+
+```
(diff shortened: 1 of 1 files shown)


oharboe:
Closing: a plan was asked for in chat, not as a PR. Superseded by one-PR-per-concern implementation.

PR #927, opened 2026-09-06
Infrastructure for A/B-testing an OpenROAD PR's QoR effect

oharboe:
Four independent pieces, least-churn-first. They stand on their own merit
whether or not any particular OpenROAD PR is ever tested again.

- **bump** openroad to `63fe72c577`, orfs to `68cc9bc97450`, so the placement a
  repair-stage measurement starts from and the code under test come from one
  revision instead of the placement lagging 63 commits behind.
- **fix(designs)**: the designs repository rule watched only the `config.mk`
  files present on the last fetch, so a design added later — by an ORFS patch,
  say — never invalidated it. `orfs_design()` then returned early on the missing
  key and declared *no targets, silently*; the error surfaced far away as
  "target not declared". Now watches the tree.
- **two carried ORFS patches**: `0050` factors `global_route.tcl`'s incremental
  repair phase into a proc a hook can stand in for, so one routed design feeds
  many repair configurations instead of re-routing per configuration (the
  extracted block's bare `return`s had to become a propagated status, or a
  failed route would have fallen through). `0051` adds `asap7/aes_multivt`,
  `aes` with the RVT/LVT/SLVT ladder and the clock retuned 380ps → 285ps —
  the retune is the point, since a faster library at an unchanged period closes
  the design and leaves a repair-stage measurement with nothing to measure.
- **`//study`**: a fork/join A/B harness. The arm is chosen per invocation via
  `OPENROAD_EXE`, so no local path is baked into a committed file. Leaves run
  ORFS's own `cts.tcl`/`global_route.tcl` on the inherited in-memory design
  rather than an imitation of them.

The harness carries three assertions, each corresponding to a bug that produced
clean, plausible, wrong numbers rather than an error — a caught-and-swallowed
stage failure, a clock constraint in the wrong units, and incomplete leaves.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -49,7 +49,7 @@ bazel_dep(name = "sv-elab", version = "2026-07-07")
 bazel_dep(name = "openroad")
 archive_override(
     module_name = "openroad",
-    integrity = "sha256-OMm/0RskoU/QGyWOnkkPjbZziplDJDAHLwF575P0AgY=",
+    integrity = "sha256-yFYlQ1zckLkO3aarUqj18Fia3au5AJgbWJvELQ4prbI=",
     # GitHub /archive/<sha>.tar.gz tarballs don't carry submodules,
     # so vendor src/sta (OpenSTA) and third-party/abc from their own
     # GitHub auto-archives at the SHAs OpenROAD's .gitmodules pins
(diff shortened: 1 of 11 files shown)


oharboe:
Closing: four concerns in one PR, and by the cruft test only two of them should merge at all.

Split into single-concern, single-commit PRs:
- the openroad/orfs bump
- the designs-repo `watch_tree` fix (a real bug: a patched-in `config.mk` could never invalidate the repo, and `orfs_design()` then declared no targets silently)

The other two — `patches/0050`, `patches/0051` and the `//study` harness — exist for one dogfooding exercise and have no in-tree user outside it. They stay in the closed reference PR #930, where they are reproducible without everyone carrying them.

PR #939, opened 2026-09-06
Study: can bazel-orfs produce the QoR data OpenROAD PR 10662 needs? (self-contained, closed for reference)

oharboe:
**OpenROAD PR 10662** — `rsz: BufferToInverters move`
https://github.com/The-OpenROAD-Project/OpenROAD/pull/10662

# Is it better?

**Yes on sky130hd, no on asap7 — and as submitted it does nothing at all on
sky130hd.**

| platform | arm | min clock period | vs base | TNS | commits |
|---|---|---|---|---|---|
| **sky130hd/aes**<br>3.6 ns target | base | 3.7869 ns | — | −3.317 ns | 0 |
| | as submitted | 3.7869 ns | **0.000%** | −3.317 ns (0.0%) | **0** |
| | review completed | 3.7270 ns | **−1.58%** | −2.362 ns (**−28.8%**) | 24 |
| **asap7/aes**<br>380 ps target | base | 384.780 ps | — | −34.50 ps | 0 |
| | as submitted | 385.352 ps | **+0.149%** | −56.30 ps (**+63.2%**) | 3 |
| | review completed | 385.352 ps | +0.149% | −56.30 ps (+63.2%) | 3 |

Minimum achievable clock period (`target − WNS`), not WNS, because WNS here is
a small number whose ratios flatter the result.

**sky130hd, with its own review applied: a 1.6% faster clock and 29% less TNS,
for +0.25% area and +7% repair runtime.** A real trade, and one worth weighing
— not transformative.

**asap7: the move fires as submitted and costs 0.15% of clock period while
making TNS 63% worse.** Note the shape: min period barely moves, TNS moves a
lot. The move is not damaging the worst path; it is damaging many
near-critical paths while barely touching the worst one — with only 3 commits.
That points at the acceptance criterion evaluating the path it sits on without
accounting for what it costs elsewhere. That is a specific question for the
author, and a more useful one than "it regressed on asap7".

**The noise floor is exactly zero.** Every `stock` / `stock_again` pair — one
configuration run as two forked leaves — agrees bit-for-bit, in all six arms,
on both platforms. The `+0.000%` rows below are measured, not rounded. So the
1.58% and the 0.149% need no confidence interval and no statistics.

## What this means for the pull request

1. **The footprint check is disabling the move on a platform where it
   demonstrably helps.** As submitted, sky130hd output is bit-identical with
   and without the move named — zero commits. The pull request's own automated
   review flagged this at high severity and a maintainer replied *"This
   feedback looks valid. It is better to remove the footprint checker for this
   move."* The check is still in the code at head `0c49c2bd`; the pull request
   was then marked stale.
2. **It should not go in the default sequence** on this evidence. It helps one
   platform and hurts the other.
3. **The TNS regression looks like an acceptance-criterion problem**, not a
   fundamental one — worth fixing before either of the above matters.

---

# Details

## Why the move cannot fire on sky130hd as submitted

ORFS sets `MATCH_CELL_FOOTPRINT=1` for the platform and forwards it into
`repair_timing`. From the leaf's own log:

```
repair_timing -setup_margin 0 -hold_margin 0 -sequence unbuffer,vt_swap,sizeup,swap,buffer,clone,split,buffer_to_inverters -repair_tns 100 -match_cell_footprint -verbose
```

In sky130hd's liberty every buffer carries `cell_footprint`
`sky130_fd_sc_hd__buf` (or `__clkbuf`) and every inverter `__inv` (or
`__clkinv`). Both are non-empty and never equal, so
`BufferToInvertersGenerator::generate` discards every candidate inverter,
`best_inv` stays null, and no candidate is ever emitted. asap7 sets no
`MATCH_CELL_FOOTPRINT`, which is why the move fires there unchanged.

`patches/openroad-10662-review-completion.patch` removes exactly that check
and nothing else. On asap7, where there is no gate to remove, it is a strict
no-op — the `as submitted` and `review completed` rows are bit-identical to
nine digits, which is the check that the patch is as narrow as it claims.

## Cell-level effect

| | sky130hd/aes | asap7/aes |
|---|---|---|
| commits | 24 | 3 |
| buffers | 4234 → 4211 (−23) | 2247 → 2233 (−14) |
| inverters | 88 → 136 (+48) | 192 → 198 (+6) |
| instances | 18383 → 18406 (+23) | 12708 → 12700 (−8) |
| area | +0.25% | −0.10% |
| repair runtime | 1401 → 1505 s (+7.4%) | 523 → 695 s (+33%) |

sky130hd's arithmetic is the move doing exactly what it says: 24 commits, each
trading one buffer for two inverters, giving −23 buffers and +48 inverters for
+23 instances.

## Three checks, all passing

**Noise floor zero** — as above, and visible in the main table.

**The binary swap took.** On `base`, the `with_move` leaf raised: that binary
does not know the move name and its log stops at the `repair_timing` line. A
base arm that had quietly accepted the sequence would mean the binary override
never took effect and every number here would be worthless.

**The null arm is clean.** `base` and `as submitted` agree bit-for-bit on
`stock`, on both platforms. The change adds a `MoveType` value and edits three
policies, so "a move nobody names changes nothing" was worth testing rather
than assuming.

## What is carried here

* `patches/openroad-10662-buffer-to-inverters.patch` — the change, **rebased
  onto the pin**. The pull request sits 2364 commits behind master, so a delta
  against its June merge base answers a question nobody is asking. The five
  commits squash and replay clean — 27 files, +732/−7 — and it builds, links
  and runs.
* `patches/openroad-10662-review-completion.patch` — ours: drops the footprint
  filter, nothing else. The reviewer's alternative, a Vt-class match, is a
  design decision for the author and maintainers, not for a study.
* `patches/0050` — the ORFS seam between routing and incremental repair that
  lets one routed design feed many repair configurations.
* `//study` — the walk, the noise-floor arm, the inertness detector.

## What is not measured

* **One design per platform.** n=1 each, and the two start in different
  regimes: sky130hd/aes is 5.19% from closure, asap7/aes 1.26%. I would not
  generalise across platforms from this.
* **Post-GR repair only.** The pull request claims improvement in post-CTS
  *and* post-GRT repair; forking at CTS forfeits the shared routing prefix and
  roughly triples the cost.
* **The `OPT_POST_GRT_WNS` pass**, which hardcodes `-sequence "vt_swap
  reroute"` — the move can never run there.
* **The other live review finding** — the candidate does not override
  `estimate()`, so the multi-threaded policy cannot rank the move. ORFS drives
  the legacy path, so fixing it would change the patch under test without
  changing what is measured.

Opened to be read and closed, not merged. The broader argument — why pull
requests go unanswered, and what a durable scoring scheme would look like —
is in #943, with its data.


oharboe:
Closed on purpose. Opened to be read, not merged: the harness, the two carried patches and the lifecycle data are one occasion, and the branch is where the occasion is carried.

The three arms have not been run yet. When they are, the numbers land here as a comment on this pull request.

oharboe:
@maliberty FYI an experiment testing bazel-orfs as AI scaffolding for studying QoR effects of an OpenROAD PR.

What's missing is data for the reviewer to decide if a PR helps or hurts. This oftentimes requires a designed study to first select suitable designs to run a pareto study on. Almost certainly the study requires PR and domain knowledge.

I think AIs can run studies that provide data on QoR to see if a PR moves the pareto front. Testing an OpenROAD PR can require ORFS changes to scripting or even a new example design or patching up config.mk or several variants of a PR.

AI nit-picking doesn't add value to my mind, it just increases reviewer and reviewee fatigue. Like formatting, it is, to my mind, best applied post-merge. A PR needs to negotiate the economical hand-off for reviewer and reviewee.

PR #947, opened 2026-09-07
demo: the .odb carries its timing constraints (carries OpenROAD #11260 + ORFS patch, with measurements)

oharboe:
# The .odb carries its timing constraints

## The pitch

OpenSTA's `read_sdc` and `write_sdc` are the human interface to constraints:
forgiving on the way in, canonical on the way out. In ORFS they are also the
transport between stages. Every substep re-reads an `.sdc` (16 reads per
flow, 6 writes), and each `.odb` is paired with its `.sdc` by globbing
`results/` and taking "the greatest `.sdc` at or below this `.odb`'s stem".
Every consumer has to remember to do that, and it has guessed wrong before.

This PR makes the `.odb` carry its constraints. `read_sdc` runs once, at
synthesis, on the file a human wrote. From there `write_db` stores the
constraints and `read_db -sdc` restores them, in C++ with no Tcl and no name
lookup. The artifact is self-contained: one file gives any build system, and
the C++ and Python APIs, a constrained design that can never be out of sync
with its netlist.

The upstream ask is The-OpenROAD-Project/OpenROAD#11260. This PR is the evidence: it carries
the whole stack against current ORFS and OpenROAD master and measures it.

- `MODULE.bazel` + `patches/0051-openroad-sdc-in-odb.patch`: the OpenROAD
  patch (The-OpenROAD-Project/OpenROAD#11260, squashed) on the openroad archive_override.
- `patches/0050-orfs-sdc-in-odb-compat.patch`: the ORFS side, step one,
  forwards compatible with any OpenROAD: `load_design` restores from the
  `.odb` when the binary can and the `.odb` has constraints, else reads the
  `.sdc` alongside as before. This is the step that can go upstream first.
- `patches/0052-orfs-sdc-in-odb-retire.patch`: step two, once the OpenROAD
  side is the floor: the fallback and the intermediate `.sdc` plumbing go.
- bazel-orfs' own stage plumbing follows.
- `test/sdc_scale/`: the measurements below, reproducible with
  `bazelisk build //test/sdc_scale:sdc_scale_200k_timing`.

## What the flow pays today

`write_sdc` has no way to say `all_registers`. The asap7 platform constraints
say `-from [all_inputs] -to [all_registers]` (and back) in three
`set_max_delay` and four `group_path` commands, so the generated `1_synth.sdc`
names every register clock pin seven times, one `[get_cells {...}]` per line.
Its size and its parse time are linear in the register count, with a large
constant: each name is a hierarchical path resolved through the Tcl layer.
The port term is quadratic (`get_ports` walks every port per call) but small
in practice. The flow then pays that parse at every one of its 16 loads.

## Measurements

Same machine, OpenROAD master 5d6f167 with and without the patch, synthesis
outputs identical. Synthetic designs are N flops of pipeline behind a 64-bit
bus with the stock asap7 platform constraints; nangate45/bp_be is the ORFS CI
design with the most ports (3029), on a platform whose constraints have no
`all_registers` term, as the real-design control. One table per question;
the raw per-step numbers are the reference tables in
The-OpenROAD-Project/OpenROAD#11260.

**Load time** (`read_db` + `read_sdc` before, `read_db -sdc` after)

| design | before | after | speedup |
|---|--:|--:|--:|
| 50k flops | 4.72 s | 0.51 s | 9.2x |
| 200k flops | 21.50 s | 2.34 s | 9.2x |
| 800k flops | 87.53 s | 10.37 s | 8.4x |
| nangate45 bp_be | 0.27 s | 0.14 s | 1.9x |

**Peak memory** (OpenSTA `memory_usage` at process end: before = `read_db`, `read_sdc`, `write_sdc`, `write_db`; after = `read_db -sdc`, `write_sdc`)

| design | before | after |
|---|--:|--:|
| 50k flops | 0.48 GB | 0.33 GB |
| 200k flops | 1.43 GB | 0.81 GB |
| 800k flops | 5.20 GB | 2.69 GB |
| nangate45 bp_be | 0.22 GB | 0.16 GB |

**Disk per flow** (ORFS writes 20 `.odb` files per run; today also 9 `.sdc` files, after only the 2 exports; synth-stage sizes stand in for every stage)

| design | before | after | net |
|---|--:|--:|--:|
| 50k flops | 1169.7 MB | 1097.7 MB | -72 MB (-6.2%) |
| 200k flops | 4652.2 MB | 4379.2 MB | -273 MB (-5.9%) |
| 800k flops | 18648.6 MB | 17623.9 MB | -1025 MB (-5.5%) |
| nangate45 bp_be | 601.2 MB | 608.8 MB | +8 MB (+1.3%) |

**`write_db` time** (the record is encoded on every write)

| design | before | after |
|---|--:|--:|
| 50k flops | 0.05 s | 0.18 s |
| 200k flops | 0.19 s | 0.70 s |
| 800k flops | 0.74 s | 2.87 s |
| nangate45 bp_be | 0.03 s | 0.10 s |

Whether the native record can carry an Sdc is decided by inspecting the Sdc
through its public accessors, so `write_db` does not write or scan any text.
One construct is not observable through OpenSTA's public Sdc API and would be
lost by the native form: `set_clock_sense -positive|-negative` (only
`-stop_propagation` is queryable). No ORFS platform or design uses it; it is
stated in the code and is the one thing an OpenSTA accessor would close.

## Fidelity

For every design above, `write_sdc` after `read_db -sdc` in a fresh process
is byte-identical to the `1_synth.sdc` the baseline flow wrote
(`cmp`, see `sdc_scale_*_restore`). Two `write_db` of the same design, and a
`write_db` after `read_db -sdc`, produce byte-identical `.odb` files
(OpenROAD `src/dbSta/test/sdc_in_db8`). Constraints the native record does not
cover fall back to the `write_sdc` text, replayed as before, so nothing is
ever dropped. QoR is unchanged by construction.

## What goes

ORFS (`patches/0050` + `patches/0052`): 29 files changed, 114 insertions(+), 164 deletions(-)

- `find_sdc_file` and its two callers (open.tcl, generate_abstract.tcl).
- `load_design`'s `.sdc` argument for `.odb` inputs.
- Seven intermediate `.sdc` files, their `write_sdc` calls, Makefile targets,
  copies and dependencies. `1_synth.sdc` and `6_final.sdc` stay as the
  human-readable exports.

bazel-orfs: 9 files changed, 43 insertions(+), 48 deletions(-)

- Stage rules and squash metadata declare no intermediate `.sdc`; the
  metadata stage no longer depends on the floorplan stage for it.

## Forward compatibility

- `read_db` without `-sdc` is unchanged.
- An OpenROAD without the patch reads the new `.odb` unchanged: the
  constraints are opaque string properties it ignores.
- `read_db -sdc` on an `.odb` without constraints does nothing, quietly;
  `ord::sdc_in_db_kind` reports `none`, so a flow can fall back to
  `read_sdc`. No existing command changes its signature or its result.
- `GUI_TIMING=0` never pays for constraints: `-sdc` is opt-in and open.tcl
  passes it only when timing is wanted.

## Retirement

Both patches retire on the `//:bump` onto an OpenROAD and ORFS that carry the
change. A secure-CI run over the ORFS design set with the patched stack is
the one thing this PR cannot produce itself.


oharboe:
just data and testing here.

oharboe:
# Feedback to act on: one payload or two?

Review feedback on the upstream PR, from @jhkim-pii
(The-OpenROAD-Project/OpenROAD#11260 (comment)):

> If usability is the primary goal, it might be better to support `sta.sdc` only.
>
> Managing both `sta.sdc` and `sta.sdc.native` increases complexity. Especially,
> `sta.sdc.native` requires continuous maintenance on OpenSTA improvements to
> retain its value, which looks error-prone.

This is a maintenance argument, and it is a fair one: the native record is a
second model of OpenSTA's `Sdc` living outside OpenSTA, and a model like that
falls behind the thing it models. Before changing the OpenROAD PR or answering
with an opinion, I want this study to carry the numbers for the alternatives,
so the choice is made on measurements rather than on taste. Reopening the
measurement question here; the upstream PR stands as is until this says
something.

## What this study already implies about `sta.sdc` only

Text-only restore *is* `read_sdc`: the same Tcl, the same name resolution, only
sourced from a block property instead of a file. So the load column reverts to
the baseline, and `write_db -sdc` gains a full `write_sdc` on every write,
which the native form does not pay. Derived from the reference tables already
in this PR (`read_db` + `read_sdc`, and `write_db` before + `write_sdc`):

**Load**

| design | today | `sta.sdc` only | native |
|---|--:|--:|--:|
| 50k flops | 4.72 s | ~4.72 s | 0.51 s |
| 200k flops | 21.50 s | ~21.50 s | 2.34 s |
| 800k flops | 87.53 s | ~87.53 s | 10.37 s |

**`write_db -sdc`**

| design | today (no `-sdc`) | `sta.sdc` only | native |
|---|--:|--:|--:|
| 50k flops | 0.05 s | ~0.78 s | 0.18 s |
| 200k flops | 0.19 s | ~3.52 s | 0.70 s |
| 800k flops | 0.74 s | ~15.66 s | 2.87 s |

So text-only is not the free simplification it looks like: it keeps the
usability win, drops the whole load win, and is ~5x *worse* than the native
form on the write side, ~15 s per `write_db` on the 800k design against 2.87 s.
Twenty writes per flow. Those two tables are arithmetic on numbers this study
already measured, not new measurements, and the first thing the re-run should
do is confirm them by actually building a text-only variant rather than
inferring it.

## Where the maintenance risk actually sits

Worth separating, because the objection names the wrong half of the code.

The encoder and decoder (~880 lines of `dbSdcInDb.cc`) are mechanical id/object
work and are pinned by a round-trip test: if they disagree, `write_sdc` before
and after stop being byte-identical and the test fails on the spot.

`Coverage` (~340 lines) is the fragile part. Its only job is deciding
native-vs-text, it does that by mirroring `WriteSdc` category by category
through `Sdc`'s public accessors, and its failure mode is silent: a construct
it does not know about is not rejected, it is simply not visited, so it is
dropped. It already has one documented blind spot in that exact shape,
`set_clock_sense -positive|-negative`, which is not observable through the
public `Sdc` API and is not detected. That is the class the objection is really
about, and it is the piece worth designing away. The text payload itself is
`write_sdc` plus `Tcl_Eval`, about six lines, and it is what makes "nothing is
ever silently dropped" true.

## The number the study is missing

There is a middle option that would collapse the two payloads into one *and*
delete `Coverage` entirely: keep `write_sdc` text as the only payload, but emit
object references as odb ids instead of hierarchical names, resolved directly
on the way back in. Every object reference `write_sdc` emits goes through five
one-line functions in `WriteSdc.cc` (`writeGetPin`, `writeGetPins1`,
`writeGetPort`, `writeGetNet`, `writeGetInstance`); everything else in the
writer only prints command syntax. Overriding those five means constructs are
never interpreted, only their object arguments are rewritten, so a construct
nobody has heard of is carried correctly by construction and there is nothing
left to keep in sync with OpenSTA.

What it removes is the part of `read_sdc` that this study showed to be the
cost: ~2.45M name resolutions and ~10 MB of the 15.5 MB of text on the 50k
design. What it keeps is Tcl parse and command dispatch. **How much of the
9.2x survives that is the number that decides the whole question**, and this
study does not have it.

It can be bracketed with no OpenROAD build at all, purely in Tcl, on the
`1_synth.sdc` files `test/sdc_scale` already produces:

1. **Parse floor** — source the expanded `1_synth.sdc` with the sdc commands
   and `get_*` shadowed by no-op procs. Pure Tcl parse and list churn, no STA,
   no name lookup. This is the best an id-referenced text form could ever do.
2. **Full read** — `read_sdc` on the same file. Already measured:
   4.49 / 20.56 / 83.59 s.
3. **Unexpanded read** — `read_sdc` on the hand-written asap7 platform
   constraints, which say `-from [all_inputs] -to [all_registers]` and never
   name a register. Same constraints, no name expansion.

(1) against (2) says how much is name resolution and therefore recoverable by
ids; (3) is the sanity check on that split from the other direction. If (1)
lands near the native record's 0.51 s, the native record has no reason to exist
and the answer to the review is "you are right, and here is a single payload
that is also fast". If (1) lands at 2 s, the native form has a real argument
and the discussion narrows to `Coverage` alone, where the fix is a census over
`write_sdc` output instead of a hand-written walk of `Sdc`.

## Re-run plan

Variants to carry, on the same designs and machine as the tables above:

- **A** baseline, today's flow (control, confirms the harness still agrees with
  the published numbers).
- **B** native record, i.e. the PR as it stands.
- **C** `sta.sdc` only, the reviewer's proposal, built rather than inferred:
  load, `write_db`, peak memory.
- **D** the three bracketing timings above, which need no patched OpenROAD.

Peak memory matters for C as much as time: the before column of the memory
table is what text replay costs, because it materialises the text and every
name in it.

Then the reply upstream, and any change to
The-OpenROAD-Project/OpenROAD#11260, follows from D.

oharboe:
# Checked: `open_synth` does no extra OpenROAD↔OpenSTA network tracking

A hypothesis worth killing before it shapes the re-run plan. Opening a stage
in the GUI is slow on large designs in a way that `load.tcl`-path stages are
not, and the natural suspicion is that `make open_synth` switches on some
extra cross-layer network tracking between odb and OpenSTA that the
`load.tcl` path never pays — which would mean `read_sdc` is not even the same
operation on the two paths, and the load numbers in this PR would not
transfer to the GUI case.

Code inspection says **no**, on both halves. Line numbers are from an
OpenROAD tree close to master and an ORFS tree close to the pinned commit;
they want re-confirming against whatever the study pins, but the structure is
core and long-standing.

## Tcl level: the same command, the same db read

- `flow/scripts/open.tcl:43` and `flow/scripts/load.tcl:42` are the same
  `log_cmd read_sdc <file>`.
- Both read the db through `read_db {*}[hier_options]` (`open.tcl:32`,
  `load.tcl:36`), and `hier_options` (`flow/scripts/util.tcl:215`) is driven
  only by `OPENROAD_HIERARCHICAL`, which `open.tcl` does not set. So there is
  no `-hier` asymmetry, and no hierarchical `dbNetwork` mode that only the
  GUI path turns on.

## Network-tracking level: identical state on both paths

- **Callback registration is path-independent.** `dbStaCbk` — the ~30-method
  `odb::dbBlockCallBackObj` at `src/dbSta/src/dbSta.cc:154` that wires every
  db edit back into STA — is attached in `dbSta::postReadDb`
  (`dbSta.cc:389-398`) and `dbSta::postReadDef` (`dbSta.cc:374-382`). Both
  paths arrive there through the same `read_db`, with the same
  `addOwner(block)` and `setNetwork(db_network_)`.
- **`dbSdcNetwork`'s path cache is lazy and path-independent.** The
  name-resolution layer every `get_cells`/`get_pins` in an SDC goes through
  builds its map on first use in `sdcPathToInstMap()`
  (`src/dbSta/src/dbSdcNetwork.cc:256`, flag at `dbSdcNetwork.hh:100`) and
  invalidates only on hierarchy edits. `read_sdc` builds it identically
  either way.
- **The GUI does register three `dbNetworkObserver`s** that exist in no batch
  session — `src/gui/src/browserWidget.h:39`,
  `src/gui/src/displayControls.h:143`, `src/gui/src/clockWidget.h:513`. But
  `dbNetwork` notifies them exactly twice, at `readDbAfter`
  (`src/dbSta/src/dbNetwork.cc:2178`) and `readLibertyAfter`
  (`dbNetwork.cc:2464`). That is a one-shot widget rebuild at open, **not** a
  per-object cost inside SDC parsing.

## The one real difference is the input, not the tracking

`open.tcl:36` picks the constraint file with `find_sdc_file` — glob
`results/` and take the greatest `.sdc` at or below this `.odb`'s stem —
where `load.tcl:42` is handed an exact path. So `open_synth` can read a
*different*, and at later stages larger, `.sdc` than the stage itself read.
That is a difference in what is read, not in how the network is tracked, and
it is the same globbing this PR sets out to delete.

## What this changes about the re-run

It removes a whole explanation, which is the useful part. GUI-open latency
cannot be hidden cross-layer bookkeeping; it has to be in the terms
`open.tcl` adds over `load.tcl`:

1. `open.tcl:71-72` — `sta::find_timing` + `sta::find_requireds` warm-up: a
   full graph build plus requireds propagation. `load.tcl` never does this.
2. `open.tcl:89-92` — `gui::select_chart "Endpoint Slack"` +
   `gui::update_timing_report`, iterating every endpoint. Guarded by
   `[gui::enabled]`, which is **false in batch** — so this term is invisible
   to `load.tcl` *and* to any batch harness that sources `open.tcl`. It needs
   a real GUI session to measure at all.
3. `open.tcl:53-68` — `set_propagated_clock`, `read_spef` /
   `estimate_parasitics`.
4. `read_sdc` itself, on whatever file `find_sdc_file` chose.

So the GUI arm of the re-run has to price (1)-(4) individually rather than
report an open-to-first-timing total, and it needs one hand-driven GUI rung
for (2), because no batch measurement can see it. If (1) and (2) dominate
(4), then the SDC payload is the wrong lever for the *GUI* complaint
specifically — which would be a finding worth having, and would leave the
case for this PR resting where it already rests: on the 16 loads per flow,
not on the GUI.

PR #948, opened 2026-09-07
study(odb): structure of arrays as a switch — evidence

oharboe:
**Evidence for making structure-of-arrays a switch in odb.** No prose file in the tree — the scripts and the CSVs are the deliverable, and this body is the argument. Opened to be citable, then closed.

## The claim

`dbTable` stores a table's fields inside its slots. Storing them as columns instead — one array per field — makes the same database smaller in two independent ways, and each is a one-line opt-in on the slot type:

```cpp
static constexpr bool kSoaTable = true;         // columns in memory
static constexpr bool kFieldMajorTable = true;  // columns on the wire
```

Measured with only `iterm_tbl` opted in, on five public ORFS designs across three PDKs:

measured

Compressed `.odb` **1.07x–1.44x** smaller (1.38x aggregate), peak RSS of `read_db` **8.8%–11.8%** lower, read and write time unchanged, uncompressed size 0.4–1.5% smaller.

## Why it works

A slot's serialized record is a row of fixed-width fields. Read as columns, most of them are constant or monotone — five of `_dbITerm`'s ten 4-byte columns hold a single value on `ariane133`, and three more have a constant first difference 65–89% of the time. Array-of-structs interleaves those with the two columns that carry real entropy, so no compressor ever sees a run longer than one field. Every record is distinct; it is the *adjacency* that changes, nothing else.

bytes per slot

In memory the same layout removes what a slot spends on padding and on fields most slots do not use: an access-point map is 24 bytes per slot and is empty on every iterm until pin access runs.

## Block size is the knob, and it is not the page size

The gain grows with run length, and a table's page can be as small as 128 slots. Raising `page_size` would give the same adjacency and must not be used — an object id is `page_addr | slot`, so it renumbers every object. A stream block spanning several pages leaves every id where it is.

gain vs block size

## Gates

Every design passes two, on every run:

- the DEF written after a round trip is **byte-identical** to the DEF written before it — no id moved, no `dbSet` order shifted, no free-list order changed;
- write → read → write reproduces the **same bytes**.

Plus three intent-based C++ tests (no goldens, per #11281 upstream): ids and connectivity survive a round trip, writing what was read reproduces it, and allocation order survives a reload — that last one because rebuilding a free list from the allocation bitmap would still be *valid* and would still renumber every later object.

## What is in the tree

`study/soa/` — `transpose_measure.py` (the block-size sweep under both policies a schema-free writer can implement), `column_structure.py` (why it works, per column), `rss_projection.py`, `read_db_cost.sh`, `roundtrip_gate.sh` (the gates and the artifact size), `io_cost.sh`, `plot.py`, the CSVs behind every number, and `patches/odb-soa-switch.patch`, which carries the OpenROAD side against the pinned revision.

## Measured, and left out on purpose

Each is a further gain the switch makes available; none is in the carried patch, so that the change stays one concern:

| | |
|---|---|
| `box_tbl` + `sbox_tbl` on the wire | 1.38x → **1.53x** |
| every table on the wire | **2.51x**, but 7.1x read time and 2.3x peak RSS — blocks are sized by `sizeof(T)`, which does not bind for records carrying a heap payload |
| dropping the derivable `oid_` | 4 bytes a slot, every table; also the field behind OpenROAD #9960 |
| dropping the derivable `prev_net_iterm_` | 7.2 MB of a 10.8 MB column total on one design |

## Two things measurement caught that reasoning did not

- A sparse column costs nothing for absent entries, but its **mutable accessor creates them** — the deserializer asking for one per iterm re-densified the column and made RSS *worse* than the dense version. The reader now stores only non-empty maps.
- After pin access, 71.8% of iterms carry an access point, so a block holds two record lengths. A whole-block transpose falls back and the gain drops 1.99x → 1.60x; splitting the block by record length recovers 1.83x for one bit per slot.

## Reproducing it

`study/soa/REPRODUCE.txt` has the command sequence: which ORFS targets build the designs, how the two probe binaries are added to an OpenROAD checkout (`study/soa/tools/`), and which script produces which CSV. It also states the two things that make peak RSS numbers comparable — subtract the ~75.7 MB the same binary uses loading nothing, and note which harness produced each baseline row.

Committed alongside: `memory-aos-baseline.csv` (the "before" column, which previously existed only in my scratch directory) and `column-census.txt` (the per-column counts behind "five of ten columns hold a single value" — worth reading, since `microwatt` has two rather than five, which is most of why its gain differs).


oharboe:
Closed on purpose: this is evidence, not a change to land. The scripts, CSVs and figures stay reachable at this URL, and the carried patch in `study/soa/patches/odb-soa-switch.patch` is what the OpenROAD-side proposal consists of.

PR #954, opened 2026-09-08
feat(bump): check ORFS_PATCHES against the bumped ORFS before writing

oharboe:
The bump follows ORFS master while bazel-orfs main may still pin an older ORFS, and bazel-orfs's own `ORFS_PATCHES` apply below any consumer patches. A consumer can therefore be moved onto an ORFS commit those patches no longer fit, and the first sign is `ctx.patch` failing inside bazel-orfs's patch dir during `mod tidy`, after `MODULE.bazel` was already rewritten. That is what happened when ORFS #4501 deleted `flow/BUILD` while main still carried 0047 and 0049 against it (fixed in #953).

This makes the bump check first: read `ORFS_PATCHES` from `orfs_source.bzl` at the bazel-orfs commit being pinned (the checkout itself when bumping bazel-orfs), verify each hunk's pre-image against the touched files of the ORFS tree at the target commit (one raw file fetch per touched file, no tarball, no `patch(1)`), and stop naming the patch, the file or hunk, and both commits. `--ignore` turns it into a warning like the other precondition checks.

Checked against the real pair: bazel-orfs a668099 vs ORFS a71b115 reports 0047 and 0049; main after #953 is clean. Unit tests cover the label parser, the diff parser (old-side counts, creations), the tree check, and the raise/warn behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,6 +378,13 @@ What it updates:
 - **bazel-orfs** git commit (latest from GitHub)
 - **OpenROAD** git commit (latest from GitHub, if configured)
 
+Before writing anything it checks that the ORFS patches bazel-orfs itself
+carries (`ORFS_PATCHES` in `orfs_source.bzl`) still fit the ORFS commit it
+is moving you to, and stops naming the patch and the commit if not -- the
+alternative is a `ctx.patch` failure inside bazel-orfs's patch dir during
+`mod tidy`, after `MODULE.bazel` was already rewritten. `--ignore` turns
+that into a warning.
+
 In downstream projects, it also injects commented-out boilerplate for
 [building OpenROAD from source](docs/openroad.md) — uncomment to test the
 latest OpenROAD before ORFS catches up. This is useful when an OpenROAD bug
--- a/bump_impl.py
+++ b/bump_impl.py
@@ -354,6 +354,183 @@ def compute_sha256_hex(url):
     return h.hexdigest()
 
 
+# --- ORFS_PATCHES pairing check ------------------------------------------
+#
+# bazel-orfs carries patches against ORFS (ORFS_PATCHES in orfs_source.bzl)
+# and applies them inside its orfs_repositories extension, below any patches
+# the consumer adds.  The bump follows ORFS master while bazel-orfs main may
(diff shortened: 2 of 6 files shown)


oharboe:
Pushed `543bb36` — grows a fake ORFS tree so the pairing check has something real to check. CI failure was 43 errors + 2 masked failures, all one cause.

**Why it was red:** the check runs before the assertions the existing tests make, and their commits are fixtures (`new_orfs_ccc333`, `new_bazel_orfs_aaa111`). Reading `orfs_source.bzl` and the patched ORFS files at those refs finds nothing, so every fixture bump refused with `orfs_source.bzl not found in bazel-orfs` and took the suite down with it. The 2 `failures` (as opposed to errors) were the legacy-shape and yosys-shape tests, where the new `BumpError` masked the refusal they assert on — both come back on their own once the check passes.

**The fixture:** `mock_fetch_raw_text` serves bazel-orfs files from the checkout (the real `orfs_source.bzl`, the real patches) and ORFS files from a tree synthesized from the patches themselves — each patched path gets a file holding that patch's hunk pre-images, so every hunk finds its context by construction. That way it follows whatever `ORFS_PATCHES` lists instead of vendoring ORFS content that goes stale at the next bump, and the real carried set is what gets parsed. Any repo other than those two raises, so a new network read in `bump()` gets caught here rather than reaching GitHub from CI.

`setUpModule` copies `orfs_source.bzl` and the patches beside the fixtures in the temp dir, since for a bazel-orfs project the check reads them from the directory holding `MODULE.bazel`. Under bazel that needs them as `data`, so `//patches:orfs_patches` is built from `ORFS_PATCHES` itself and cannot drift from the list that is actually applied.

**Three tests on the wiring:** the carried patches fit the tree; a drifted tree makes the bump refuse, naming the patch and the commit, and leaves `MODULE.bazel` byte-identical; and the tree holds every patched path — that last one because a fixture that silently synthesized nothing would leave the check inert with the suite still green. Emptying the tree fails 48 of the 122 tests.

122/122 pass under bazel, 69/69 across the non-ORFS unit set, `fix_lint` / `public_surface` / `host_tools` clean. Left the check's own ordering and semantics alone.

PR #958, opened 2026-09-08
study: pre-route pessimism, and what asap7 timing policy was fitted to

oharboe:
This is a **documentation PR**: it is here to be read, then closed. Nothing
in it merges. The parts worth keeping are being extracted into their own PRs
(#959 merged, #960 open); what remains here is the evidence.

Every table and figure is generated from result JSONs in
`test/pre_route_pessimism/results/` by
`bazelisk run //test/pre_route_pessimism:report` and `:plots`, so any claim
can be re-derived rather than taken on trust.

## The three findings

**1. A repair knob that costs nothing and saves half a stage.** Going from
`TNS_END_PERCENT=100` (the ORFS default: fix every violating endpoint) to 0
(fix only the worst path) costs **0.17 ps** of achieved period against a
1.44 ps resolution, and saves **135 of 270 seconds** of global route. Twelve
seeds per arm. It is the one effect here that clears its own noise floor.

**2. Resistance transfers between designs; capacitance does not.** Running
ORFS's own fit on a design sharing nothing with the four asap7 was fitted
from reproduces every per-layer **resistance** exactly (ratio 1.0000, R² =
1.0000) and misses **capacitance** by −25% to +12%. Resistance per unit
length is geometry and material; capacitance depends on what sits beside the
wire. So a single platform RC table is half right by construction.

**3. Global route is the noisy end, not the estimate.** Over twelve
placement seeds, `min_period` at grt has a 2σ spread of 13.8% against 4.3%
at placement — the routed number scatters 3.6× more than the pre-route
estimate that everything downstream distrusts. One run per arm resolves
nothing below ~15 points of place-to-grt gap.

## Two things this does NOT show

The premise was that placement reads **pessimistic** against global route by
tens of percent. On a public design with the stock platform RC it reads
**optimistic by 12%**, and only when a routing layer is over capacity. A
pessimistic reading is reachable by raising `set_wire_rc`, but that is a
choice about an input, not a property of the flow.

Nothing here says coverage never matters. `repair_timing` stops at WNS zero,
so at any clock repair can reach, every setting lands on the same floor and
the period cannot separate them. The period null is a bound at that clock,
on this design.

## The regime ORFS has no design in

asap7 ships a 9-layer stack (M1, M2, M3, M4, M5, M6, M7, M8, M9) and caps signal routing at **M7** by platform default. Of 24 asap7 design entries in ORFS, **1** raises that cap above the default (`aes-block`).

The last column is the point. Every pre-route stage prices every net at one resistance constant, so a design's longest nets -- the ones that would route high -- are charged that constant instead of the layer they would get. The factor is how many times more resistive the constant is than the least resistive layer the router could have given them.

| design | MAX_ROUTING_LAYER | set by | signal wire mispriced | timing policy it overrides |
| --- | --- | --- | --- | --- |
| `aes` | M7 | platform | 2.24x | `TNS_END_PERCENT=100` |
| `aes-block` | M9 | design | 3.15x | `HOLD_SLACK_MARGIN=-30` |
| `aes-block_aes_rcon` | M5 | design | 1.47x | - |
| `aes-block_aes_sbox` | M5 | design | 1.47x | - |
| `aes-mbff` | M7 | platform | 2.24x | `TNS_END_PERCENT=100` |
| `aes_lvt` | M7 | platform | 2.24x | `RECOVER_POWER=100`, `TNS_END_PERCENT=100` |
| `coralnpu` | M7 | platform | 2.24x | - |
| `cva6` | M7 | platform | 2.24x | - |
| `ethmac` | M7 | platform | 2.24x | `HOLD_SLACK_MARGIN=-10` |
| `ethmac_lvt` | M7 | platform | 2.24x | `RECOVER_POWER=1` |
| `gcd` | M7 | platform | 2.24x | `SKIP_LAST_GASP=1` |
| `gcd-ccs` | M7 | platform | 2.24x | `SKIP_LAST_GASP=1` |
| `ibex` | M7 | platform | 2.24x | `TNS_END_PERCENT=100` |
| `jpeg` | M7 | platform | 2.24x | `TNS_END_PERCENT=100` |
| `jpeg_lvt` | M7 | platform | 2.24x | `RECOVER_POWER=100`, `TNS_END_PERCENT=100` |
| `minimal` | M7 | platform | 2.24x | `REMOVE_ABC_BUFFERS=1`, `SKIP_CTS_REPAIR_TIMING=1`, `SKIP_INCREMENTAL_REPAIR=1` |
| `mock-alu` | M7 | platform | 2.24x | - |
| `mock-cpu` | M7 | platform | 2.24x | `TNS_END_PERCENT=100` |
| `riscv32i` | M7 | platform | 2.24x | `TNS_END_PERCENT=100` |
| `riscv32i-mock-sram` | M7 | platform | 2.24x | `TNS_END_PERCENT=100` |
| `riscv32i-mock-sram_fakeram7_256x32` | M4 | design | 1.47x | - |
| `swerv_wrapper` | M7 | platform | 2.24x | - |
| `tinyRocket` | M7 | platform | 2.24x | `TNS_END_PERCENT=100` |
| `uart` | M7 | platform | 2.24x | `SKIP_GATE_CLONING=1`, `TNS_END_PERCENT=100` |

## One number, four stages, and an inversion

`TNS_END_PERCENT` defaults to **100** and is declared for stages **place, cts, floorplan, grt**. It has exactly one functional call site: the helper every repair stage goes through appends it as `-repair_tns`.

| site | code |
| --- | --- |
| `global_route.tcl:122` | `[list -setup -sequence "vt_swap reroute" -skip_last_gasp -repair_tns 0 -verbose]` |
| `util.tcl:35` | `append_env_var additional_args TNS_END_PERCENT -repair_tns 1` |

So one number governs repair at 4 stages that know very different amounts about the design:

| stage script | line |
| --- | --- |
| `cts.tcl` | 67 |
| `floorplan.tcl` | 159 |
| `global_route.tcl` | 101 |
| `resize.tcl` | 32 |

And the one pass that runs *after* global routing -- the only point in the flow where the timing picture is route-aware -- bypasses that helper and hardcodes `-repair_tns 0`. The policy is applied where the estimate is worst and switched off where it is best.

## The two PDKs do not do the same thing

`set_wire_rc` is what turns a pre-route wirelength into a delay, and the two platforms in tree disagree structurally about it. One names a layer and inherits its resistance; the other pins a constant that no longer tracks the stack it came from.

| platform | net type | how it prices a wire | cost of that choice | spread over routable layers | spread over full stack |
| --- | --- | --- | --- | --- | --- |
| `asap7` | signal | absolute `R=2.65684E-02` | 2.24x more resistive than `M6` | 2.63x | 8.34x |
| `asap7` | clock | absolute `R=2.14161E-02` | 1.80x more resistive than `M6` | 2.63x | 8.34x |
| `sky130hd` | signal | names layer `met1` | tracks that layer | 66.54x | 3909.39x |
| `sky130hd` | clock | names layer `met3` | tracks that layer | 66.54x | 3909.39x |

This is the sharpest candidate for moving a default into the PDK: one platform already does it the other way.

## Sizing: the smallest design that still shows it

The design has to fit inside an OpenROAD edit/measure loop and still be long enough that wire delay, not gate delay, sets the period. asap7 makes that tension sharp: its cells are small enough that a die big enough for a long wire needs a lot of instances. So the size is a measured choice.

| GROUPS | std cells | core um | utilization | synth..grt | slowest stage | which |
| --- | --- | --- | --- | --- | --- | --- |
| 32 | 14,005 | 63 x 62 | 60.6% | 312 s | 176 s | 1_1_yosys_canonicalize |
| 64 | 27,023 | 88 x 88 | 58.6% | 1066 s | 848 s | 1_1_yosys_canonicalize |

## Shape: buying wire length without buying runtime

Wire delay needs physical distance, and the size curve above shows distance is the expensive way to get it. Utilization buys it at constant instance count and constant runtime. This is not a trick to flatter the benchmark: a design whose die is set by something other than this logic runs at low core utilization as a matter of course, and large designs do.

`PLACE_DENSITY` follows `CORE_UTILIZATION` down at every point. Left at the platform default it lets global placement pack every cell into a corner of a larger die and hand back short wires, so the die would grow and the measurement would not move.

| shape | utilization | core | synth..grt | demand on top metal | most contended layer |
| --- | --- | --- | --- | --- | --- |
| `u12` | 13.5% | 134 um | 193 s | 10.4% on M8+M9 | M2 26.3% |
| `u12_tight` | 13.6% | 134 um | 269 s | 10.6% on M8+M9 | M2 32.7% |
| `u25` | 27.5% | 93 um | 174 s | 9.0% on M8+M9 | M2 40.0% |
| `u55` | 60.6% | 63 um | 300 s | 6.8% on M8+M9 | M2 63.9% |
| `u55_derated` | 60.4% | 63 um | 151 s | 9.3% on M8+M9 | M2 114.2% |
| `u80_dense` | 86.5% | 52 um | 170 s | 8.2% on M8+M9 | M2 80.4% |

And the mechanism, rather than the permission. A guide on the top layer is far longer than one on M2, which is what it means for the long nets to go up -- and those are exactly the nets a single resistance constant misprices:

| shape | mean guide on M2 | mean guide on M9 | ratio |
| --- | --- | --- | --- |
| `u12` | 2.02 um | 9.72 um | 4.8x |
| `u12_tight` | 1.69 um | 7.91 um | 4.7x |
| `u25` | 1.56 um | 8.05 um | 5.2x |
| `u55` | 1.19 um | 9.12 um | 7.7x |
| `u55_derated` | 0.93 um | 8.77 um | 9.4x |
| `u80_dense` | 1.03 um | 8.31 um | 8.0x |

## The gap across the shape axis

`min_period = clk_period - WNS` per stage, per shape. The last column is the study's headline: how far the reading global placement hands downstream sits from the one global routing produces. Positive means the pre-route reading is pessimistic.

Every repair budget between those two columns -- `TNS_END_PERCENT`, `SETUP_SLACK_MARGIN`, and which stages are allowed to act -- is set against the left-hand number.

**6 shapes close at every stage** (`u12`, `u12_tight`, `u25`, `u55`, `u55_derated`, `u80_dense`). `repair_timing` stops at WNS zero, so at this clock every setting of a repair knob reaches the same floor and the achieved period cannot separate them -- what differs is runtime. Deltas below are min_period in ps, never percentages of WNS: WNS is a small number near closure and a percentage of it says more about the clock than about the flow. The policy arms use a tighter clock, chosen so repair still has work left when it stops.

| shape | `2_floorplan` | `3_place` | `4_cts` | `5_1_grt` | place vs grt |
| --- | --- | --- | --- | --- | --- |
| `u12` | 214.5 | 398.5 | 399.8 | 407 | **-2.1%** |
| `u12_tight` | 214.5 | 398.5 | 399.8 | 388.1 | **+2.7%** |
| `u25` | 214.5 | 341.8 | 344 | 346.8 | **-1.4%** |
| `u55` | 214.5 | 297.7 | 306.7 | 305.3 | **-2.5%** |
| `u55_derated` | 214.5 | 291.6 | 294.6 | 322.3 | **-9.5%** |
| `u80_dense` | 214.5 | 286.2 | 282.2 | 286.8 | **-0.2%** |

## What a single run is worth

The same configuration, 12 values of `GPL_RANDOM_SEED`, nothing else changed. This is not an error bar on the measurement -- it is the flow's own dispersion, and it bounds what any comparison on this design can claim.

| quantity | mean | 2 sigma | relative | range |
| --- | --- | --- | --- | --- |
| `min_period` at place | 302.95 ps | 13.16 ps | 4.3% | 291.6 .. 315.9 |
| `min_period` at grt | 345.46 ps | 47.53 ps | 13.8% | 317.6 .. 387.3 |
| place-to-grt gap | -12.00 % | 10.31 % | 10.3 pts | -20.1 .. -6.2 |

**Global route is the noisy end.** Its `min_period` scatters 3.6x more than placement's, which is the reverse of how the flow treats them: the pre-route estimate is the reproducible number and the routed one is the draw.

**And it sets the price of an answer.** The resolvable difference at `k` runs per arm is `2 sigma sqrt(2/k)`, so one run per arm distinguishes nothing smaller than **14.6 points** of gap, and 12 runs per arm brings that to 4.2. Any knob whose effect is below the first number has not been measured, however clean its table looks.

## Does TNS_END_PERCENT earn its runtime?

`TNS_END_PERCENT` defaults to 100 -- fix every violating endpoint -- reaches four stages through one call site, and no asap7 design lowers it. Twelve seeds per arm at a 265 ps clock, chosen so repair still has work left when it stops.

| TNS_END_PERCENT | min_period mean (ps) | 2 sigma | grt stage mean (s) | 2 sigma |
| --- | --- | --- | --- | --- |
| `0` | 266.13 | 3.41 | 135 | 86 |
| `10` | 265.88 | 2.52 | 281 | 192 |
| `100` (default) | 266.30 | 3.64 | 271 | 215 |

| arm | quantity | delta vs default | resolvable at 12 seeds | verdict |
| --- | --- | --- | --- | --- |
| `TNS=0` | min_period | -0.17 ps | 1.44 ps | did not resolve |
| `TNS=0` | grt runtime | -135.34 s | 66.93 s | **resolved** |
| `TNS=10` | min_period | -0.42 ps | 1.28 ps | did not resolve |
| `TNS=10` | grt runtime | +10.71 s | 83.16 s | did not resolve |

**Fixing only the worst path costs nothing measurable in achieved period and saves half the global-route stage.** That is the one effect in this study that clears its own noise floor.

The period null is a bound, not an absence: it is below 1.5 ps against a mean near 266. And it is what the calibration predicted -- `repair_timing` stops at WNS zero, so at any clock repair can reach, every setting lands on the same floor and the period cannot separate them. Read the period column as "coverage did not buy period *here*", not as "coverage never matters".

`TNS_END_PERCENT=10`, the value the ORFS docs suggest for runtime, is indistinguishable from 100 in **both** columns on this design. Only going to 0 buys anything.

## Is the gap a property of the flow, or of one constant?

`set_wire_rc` installs the single resistance and capacitance every pre-route stage prices a net with. asap7 ships an absolute constant; sky130hd names a layer. These arms name a layer instead, one per arm, changing exactly that line and nothing else.

| wire_rc layer | place | grt | place vs grt |
| --- | --- | --- | --- |
| `M2` | 306.3 | 371.6 | -17.6% |
| `M4` | 298.8 | 362.6 | -17.6% |
| `M6` | 303.8 | 323 | -6.0% |
| `M8` | 267.6 | 328.3 | -18.5% |

**Verdict: did not resolve.** The arms span 12.5 points, against a single-run resolution of 14.6 points from the seed ensemble. One run per arm cannot separate them, so this says nothing about the constant either way -- and reporting the spread as an effect would be the error this study set out to document.

## What actually transfers between designs

The same procedure ORFS uses for the platform -- `write_rc` then `correlate_rc` -- run on this design, and the result divided by the values the platform ships. A ratio of 1 means the fit reproduced the platform exactly.

| layer | resistance ratio | capacitance ratio | res R2 | cap R2 |
| --- | --- | --- | --- | --- |
| `M2` | 1.0000 | 1.0960 | 1.0000 | 0.9728 |
| `M3` | 1.0000 | 1.0055 | 1.0000 | 0.9780 |
| `M4` | 1.0000 | 0.9910 | 1.0000 | 0.9768 |
| `M5` | 1.0000 | 0.7949 | 1.0000 | 0.9422 |
| `M6` | 0.9999 | 0.8547 | 1.0000 | 0.9708 |
| `M7` | 1.0000 | 0.7448 | 1.0000 | 0.9745 |
| `M8` | 0.9999 | 1.1125 | 1.0000 | 0.9357 |
| `M9` | 0.9999 | 1.1212 | 1.0000 | 0.8991 |

**Resistance reproduces the platform on every layer**, on a design sharing nothing with the four the platform was fitted from, with an R-squared of 1.0000 throughout. Resistance per unit length is geometry and material: it is a property of the technology and it transfers.

**Capacitance does not**, and it fits worse. Capacitance per unit length depends on what sits beside the wire -- local routing density -- which is a property of the design, not the stack.

| wire_rc blend | resistance | capacitance |
| --- | --- | --- |
| `signal` | -5.2% | -1.8% |
| `clock` | -6.0% | +3.0% |

So a single platform RC table is half right by construction. The per-layer resistances are shared and worth shipping; the capacitances and the blend that averages them are design-specific, and a design far from the fitting population inherits somebody else's.

## Figures

min_period by stage, one line per shape

*min_period by stage, one line per shape*

the place-to-grt gap against how contended the worst layer is

*the place-to-grt gap against how contended the worst layer is*

routing demand per layer against what the router may spend

*routing demand per layer against what the router may spend*

the knob against the noise, both columns

*the knob against the noise, both columns*

what transfers between designs: resistance yes, capacitance no

*what transfers between designs: resistance yes, capacitance no*

Regenerated by `bazelisk run //test/pre_route_pessimism:plots` from the same JSONs as the tables. Three palette slots sit below 3:1 on a light surface, which is why every figure ships beside its table.


oharboe:
@maliberty FYI, bazel-orfs now has the infrastructure to run the scripts that examine final RCX to generate the setRC.tcl files.

Issue #970, opened 2026-09-10
study: thread-count idempotency of OpenSTA/rsz across asap7 and sky130hd, with a CI invariance test


oharboe:
## Why

The `-threads` scaling study (#968) is the ground under OpenROAD#11368 (gpl capped at the physical core count) and its planned follow-up (OpenSTA capped the same way, the rsz/repair_timing share of what is left). Both caps only lower the thread count, so they are safe. But the study also found a claim it could not fully back: **every substep produced byte-identical output across 2 to 32 threads on 6 asap7 designs (372 samples), yet no arm ran at 1 thread, and nothing in OpenROAD or OpenSTA CI tests thread-count invariance of STA at all.** OpenROAD regressions run single-threaded; gpl has `mt_invariance01`, STA has nothing.

The history says this matters. Thread-count non-idempotency in STA has been found and fixed at least four times, each time by a flow user diffing outputs, never by a test:

| report | shape | fix |
| --- | --- | --- |
| ORFS#3046 (2025-04, gf180/aes) | different TNS at 1 vs N threads, TSAN clean, diverges at rsz iteration 112 | OpenSTA 7e5cf132: per-vertex path index was an unordered map keyed by `Tag*`; pointer order depends on allocation order, which depends on scheduling |
| ORFS#3180 (2025-05, CTS) | crash only with threads | OpenSTA 849426fc |
| OpenROAD#9781 (2026-03, gcd) | repair_timing hash differs run to run at 16 threads, identical at 1 | rsz driver-level cache not invalidated after buffering, exposed by parallel BFS |
| OpenROAD fork Feb 2026 | multi-mode merge forced STA to 1 thread "until issues resolved"; lifted 4 days later on "crashes are rare or none" | OpenSTA fe31eef7 restored stable_sort because `worstSlack()` tie-breaking had changed |

Plus a steady stream of upstream race fixes in the same shape (shared scratch state touched inside parallel BFS visitors: cc9eb1f1, 1258bdc2, f54ab5b1, b51885aa, b4be3c53, 30c8230f, and 1d4b5162 in March 2026, a silent wrong-delay race). Upstream deleted `hashPtr` in May 2026 with the comment "pointer hashing causes results to change from run to run; use Network::id functions instead". Seventeen pointer-keyed unordered containers remain in non-test OpenSTA code.

[...]

PR #968, opened 2026-09-10
FYI: how the flow's parallel regions scale, and what -threads should be

oharboe:
# How the flow's parallel regions scale, and what `-threads` should be

**FYI only -- this PR is for the data and will be closed.** No flow
behaviour changes here, and it recommends *no* change to ORFS.

## TL;DR -- what to do after the global-placement PR

[OpenROAD#11368](https://github.com/The-OpenROAD-Project/OpenROAD/pull/11368) caps **global placement** at the physical
core count, in the tool, per region, with `-threads` left as the
ceiling. That is the right shape and the precedent to copy.

On the evidence below -- 106s of 1295s available across 6 asap7 designs -- what is left after it:

| do this next | worth | vs gpl |
| --- | --: | --: |
| **rsz + OpenSTA** -- cap at ~the core count | 85s (80% of what is left) | 4.4x |

and then **stop**. Everything else together is 2s across 4 substeps, which is smaller than the run-to-run spread of one large design -- there is no third thing worth doing on this evidence.

### Do not do these

- **Do not cap `drt`** (detailed route, pin access). It wants the
  ceiling; capping it at the core count costs about 15%.
- **Do not lower ORFS's `NUM_CORES` default.** It is the ceiling,
  not a target. Deriving it from physical cores is wall-neutral
  overall *and* would cap `drt`.
- **Do not "tune grt".** FastRoute is a few percent of `5_1_grt`;
  that stage's win is `repair_timing`, which is rsz and OpenSTA.

### Two things to know before scheduling the next one

- It is **not** a five-line mirror of #11368. `placementThreads()`
  caps one tool's own OpenMP sites; OpenSTA's thread count is set
  once globally (`OpenRoad::setThreadCount` -> `sta_->setThreadCount`)
  and shared by every consumer, so scoping it to the repair calls
  needs a set/restore around them or an STA-side cap.
- **gpl's optimum moves with design size** on this set, and STA's
  may too. #11368 measures a 1.28M-instance design preferring the
  core count while these (much smaller) designs prefer half of it --
  the same curve sampled further along. That argues for a
  work-per-thread heuristic rather than a constant.

---

*Everything below is reference detail: the per-stage and per-region
ladders, the phase attribution, and the raw samples in a comment.
It is here to be read by a machine later, not reviewed now.*

`NUM_CORES` -> `openroad -threads N` is a **ceiling**: what a job is
permitted to use, not a target. So the question is not "cores or
hardware threads" but, per parallel region, how far *below* the
[...]

PR #979, opened 2026-09-12
feat(threads): the -threads idempotency harness, and the STA invariance regression

oharboe:
The harness that answered #970, plus the regression it says was always missing.

This is the infrastructure half, proposed for merge. The findings go on #970 itself.

## Why this merges instead of dying with a study branch

#968 built a `-threads` harness, published its numbers, and closed. The branch was never merged, so answering #970 started by cherry-picking five commits off `origin/study/threads-policy`. Left alone, the next person re-cherry-picks mine. The `study-pr` skill says it plainly: *anything durable is split into its own PR first.* This is that PR.

The first five commits are #968's, unchanged, with authorship intact.

## What it does

`bazelisk run //test/threads_policy:campaign` runs one arm — a stage's substeps at one thread count — and records what it cost and what it produced. Two modes, because the two questions want opposite machines:

| mode | machine | answers |
| --- | --- | --- |
| `timing` | one arm at a time, idle asserted at every arm start | how long does it take |
| `idempotency` | arms concurrently, each in its own `FLOW_VARIANT`, no idle gate | did they compute the same thing |

The second mode is the reason the campaign finished in hours instead of days: whether two arms agree does not depend on how busy the machine was, and the scheduling perturbation is a *feature* there — contention is what exposes an order-dependent bug. Its wall times are recorded, flagged `contended`, and refused by every ladder in `report.py`.

That refusal is load-bearing. Analysing the contended samples by hand produced a confident "route knees at t=4" that the idle-gated run later showed to be pure artifact: at `--jobs 2`, t=4 is 8 threads on 8 cores, so every higher rung was oversubscribed by construction.

## The parts

| file | what it is |
| --- | --- |
| `campaign.py` | runs arms; resumable, one JSON per sample; refuses a thread count the host cannot install |
| `deployment.py` | where one arm's files are, as opposed to another's |
| `witness.py` | the three things that prove two arms produced the same result |
| `idempotency.py` | run-to-run nondeterminism vs thread-dependence — different bugs, different fixes |
| `watchdog.py` | turns a hang into a backtrace instead of a missing row |
| `elapsed.py`, `phases.py` | #968's log readers, unchanged |
| `report.py` | generates the write-up from whatever the campaign recorded |
| `sta_audit.py` | the static class-1 audit; needs no flow runs |
| `mt_invariance.tcl` | the regression, below |

Nine unit-test targets, all non-manual. Every flow-running target is `manual`.

## The regression

`//test:lb_32x128_mt_invariance_test` runs `repair_timing` at 1 and 8 threads and byte-compares the metrics report and the post-repair `.odb`, with the comparator `gcd_single_flow` already uses.

Thread-count non-idempotency in OpenSTA has been found and fixed at least four times — ORFS#3046, ORFS#3180, OpenROAD#9781, the Feb-2026 fork fence — every time by a flow user diffing outputs, never by a test. OpenROAD's regressions run single-threaded; gpl has `mt_invariance01`, STA has nothing. This is the missing one, in the shape gpl's already has.

Its `BUILD` comment states what it does not cover: `lb_32x128` is small and STA's parallel regions barely engage, so it guards against a regression reaching this repo rather than proving invariance in general.

## What it produced

400 arms — 10 designs, 3 PDKs, 4 stages, t=1…16, two repeats, three witnesses per substep. **250 of 250 verdicts stable.** No thread-dependent divergence, no run-to-run nondeterminism, including on `nangate45/gcd`, the OpenROAD#9781 reproducer.

## The witness is the part that took the most rework

An `.odb` hash alone misses a divergence in both directions, so there are three: `.sdc` bytes, `.odb` bytes, and the comparable subset of ORFS's own metrics. Three bugs in that subset were found by looking at real samples, and each is now a test:

- `5_2_route` compared **one** metric key and `5_3_fillcell` compared **none**, both reporting `stable`. Route runs no STA, so its metrics are drt's counts; and `5_3_fillcell.json` writes a bare `design__violations` with no stage prefix, which suffix matching silently dropped.
- An empty QoR dict compared equal across every arm — "identical" backed by zero comparisons.
- A witness a stage *cannot* have (route writes no `.sdc` from the substeps an arm runs) is now distinguished from one that went missing. The first gets no verdict; the second is `unproven`.

This is #968's documented failure in the other direction: there, comparing everything produced a fake divergence across 48 of 54 pairs; here, comparing nothing produced a fake agreement.

`unproven` is a verdict, not a silence — a missing witness, a ladder with no t=1 reference, or a ladder holding only the reference are all counted apart from `stable`.

## Two more the harness's own tests caught

- The class-1 audit reported six "high risk" OpenSTA sites; three were a local variable's name matching a namesake in another file. Locals are now confined to their declaring file. It also called `Sim::clearSimValues` a defect on the strength of the loop alone — following it shows the observer only invalidates, into an id-ordered `std::set`. That claim would otherwise have gone into a pull request.
- The watchdog filed gdb's refusal as if it were the stack. Yama `ptrace_scope=1` lets `gdb -p` attach only to a descendant, and gdb exits **0** while refusing, so the first real hang produced 288 bytes of error text. Frames now decide, with SIGABRT as the fallback, and the campaign warns at startup when the host cannot dump a hang.

## Checks

`//:fix_lint`, `//:public_surface`, `//:host_tools` clean. `bazelisk test //test/threads_policy:all` green. `bazelisk query 'attr(tags, manual, //test/threads_policy:*)'` shows every flow-running target manual.


oharboe:
Closed as intended — reference material, not a merge candidate.

Everything stays reachable on `feat/threads-idempotency-harness`, which is not deleted: the campaign harness (`test/threads_policy/`, 9 unit-test targets), the class-1 static audit, the report generator, and the `mt_invariance` probe. Its last commit, `5bc2eb1`, postdates the close and carries two corrections to the write-up.

**The probe does not belong in this repo.** A guard in bazel-orfs CI does not protect the people who would break it — an OpenROAD change lands against green upstream CI and surfaces downstream at bump time, which is the pattern #970 was opened about. Its home is OpenROAD's regression suite beside `src/gpl/test/mt_invariance01`, and it is listed there as an upstream candidate, not acted on. The working version here is a reference implementation for whoever takes that up.

Findings: https://github.com/The-OpenROAD-Project/bazel-orfs/issues/970#issuecomment-5644470462

The durable fixes this campaign turned up were split out and merged on their own: #971, #972, #973, #974, #975.

PR #981, opened 2026-09-13
study: ORFS already ran the noise experiment — run many cheap designs, not a curated few

oharboe:
**This is a documentation PR: it is here to be read, then closed.** It
carries a harness (`test/qor_noise/`), the data it produced
(`docs/studies/qor-noise/`) and five figures. Nothing here is proposed for
merge. No upstream repository was written to.

---

## ORFS already ran the noise experiment. Here is what it said.

Every metric in a `rules-base.json` traces back to METRICS. Fenstermaker,
George, Kahng, Mantik and Thielges proposed the infrastructure at DAC 2000,
on the precept that measuring a design process is a prerequisite to
optimising it; METRICS 2.0 (Hashemi, Ho, Kahng, Liu and Reda, WOSET 2018)
revisited it for machine learning; and METRICS2.1 (Jung, Kahng, Kim and
Varadarajan, ICCAD 2021) fixed the hierarchical
`stage__category__name__modifier` naming that `finish__design__instance__area`
is written in. When CI compares a run against `rules-base.json`, it is asking
a METRICS2.1 question.

There is a second paper, twenty years older, that CI is *not* asking. In
*Measurement of Inherent Noise in EDA Tools* (ISQED 2002), Kahng and Mantik
proposed behaviour criteria for tools — **monotonicity**, **smoothness**,
**scaling** — and a taxonomy of perturbations that change no legal property
of the problem: seeds and RNGs, ordering and naming, library richness,
constraints, and geometry such as a shifted site or row. Then they measured
it, and the ranking is the part worth carrying into 2026:

| perturbation | measured effect on solution quality |
| --- | --- |
| re-ordering the netlist | **no effect at all** — tools canonicalise on read |
| re-seeding, 200 runs | **~±0.2%** |
| randomising net names | **>7% spread** |
| scrambling name hierarchy | **up to 12% worse, 5% on average** |

The loudest noise source is not the random seed. It is *the netlist looking
slightly different* — which is exactly what every resizer, CTS and buffering
change does on its way to a legitimately better result. A PR that inserts
differently-named buffers is drawing from the 7% urn, not the 0.2% urn, and
a CI table of one run per design cannot tell the two apart. Their ICCAD 2000
companion, *On Mismatches Between Incremental Optimizers and Instance
Perturbations in Physical Design Tools*, is the same finding from the
incremental side; Chan, Kahng and Woo revisited it at SLIP 2020 and still
found **>10% variation in routed wirelength** with modern tools. METRICS2.1
[...]

PR #977, opened 2026-09-11
study: a placement seed moves global-route QoR by ~1% of the clock, and ORFS's TNS regression guards are up to 59x tighter than that noise

oharboe:
**This is a documentation PR: it is here to be read, then closed.** It carries the harness, the corpus snapshot, the figures and the raw data for one question -- what a placement seed is worth at global route, and whether the spread it produces hides a bug. Nothing here is proposed for merge; the harness is on this branch only.

## The finding

Changing nothing but `GPL_RANDOM_SEED` moves `min_period` at global route by **11.4 ps, 2 sigma 2.08% of the clock**, on jpeg; every design measured lands between 0.46% and 2.08%.

That spread is the perturbation's and nothing else's: `-perturb_dist 0` collapses aes (8 seeds), ethmac (8 seeds), gcd (8 seeds), ibex (8 seeds), jpeg (8 seeds), uart (8 seeds) to a single value across every seed. Any other seeded randomness anywhere in place, cts or global route would have shown up here as a spread, and none did. It is the control that makes the rest of the numbers mean anything.

Against ORFS's own accepted QoR -- 79 `rules-base.json` files, inverted and fitted -- that spread is small: designs sit a robust sigma of **0.128 of a clock period** from what their size and platform predict, so placement luck accounts for well under a percent of the variance between two designs in the corpus. The seed is not what makes one design's QoR different from another's.

One consequence is immediate and is not about placement at all. ORFS pads a global-route timing rule by a fixed fraction of the clock and generates it from a single run. The WNS margin (5%) covers the seed spread everywhere it was measured. The TNS margin (20%) does not: on aes, ethmac, gcd, ibex, jpeg, uart the seed-to-seed spread is 5.0x and 1.9x and 3.4x and 1.2x and 59.2x and 9.8x the margin, and a rule written from the median seed rejects 18 of the runs that differ from it by nothing but `GPL_RANDOM_SEED`. Those guards encode which seed the maintainer drew.

What the ensemble *is* good for is pricing a change, which no single run can do: below the noise floor a single run cannot tell a real gain from which seed it drew. What it priced here, each against the same design's base arm and each quoted with the resolution at the seeds run:

| what | design | min_period delta | resolution | verdict |
| --- | --- | ---: | ---: | --- |
| a 5x smaller perturbation radius (100 nm) | aes | -0.83 ps | 2.31 | did not resolve |
| a 4x larger perturbation radius (2000 nm) | aes | +3.38 ps | 2.87 | resolved |
[...]

PR #985, opened 2026-09-14
study: where global placement starts its cells is worth nothing to the finished design, and 0.40x the seed's own noise

oharboe:
**This is a documentation PR: it is here to be read, then closed.** It carries
one instrumentation patch (`patches/0066-…`, study branch only), a harness
(`test/gpl_initial_position/`), the data it produced
(`docs/studies/gpl-initial-position/`) and three figures. Nothing here is
proposed for merge. No upstream repository was written to.

---

## Where global placement starts from, and what that is worth

Global placement puts every placeable instance somewhere before it solves
anything. `InitialPlace::placeInstsInitialPositions()` picks a region centre,
else the ODB location, else the core centre — and which of the last two you get
depends on a flag.

Nobody had measured it. `placeInstsCenter()` arrived with the RePlAce import
(`456cde4187`, 2020-08-11) carrying the comment *"normally, initial place will
place all cells in the centers"* — no citation, no test. The policy has been
reversed four times since, and **not one of those changes has a QoR number
attached to it**:

| when | change | stated reason |
| --- | --- | --- |
| 2024-06 | OpenROAD #5253, use the ODB position if present | PR body empty |
| 2024-07 | #5253 follow-up, ODB only when skipping initial place | "you might pick up the skip_io result" |
| 2025-06 | OpenROAD #7644, back to the core centre | "the **expected** behavior … is the center"; "I do **not expect** any functional change" |
| 2026-01 | OpenROAD #9353, ODB by default, add `-force_center_initial_place` | "some users … **expect**"; "which is **non intuitive**" |
| 2026-01 | ORFS `bab0134ebd`, pin `-force_center_initial_place` flow-wide | "maintain default behavior" |

The one time evidence spoke — #9353's unit tests changed results — the tests
were updated.

This study runs the experiment: nine named starting distributions against ORFS
as it stands, 13 designs, 6 platforms.

## TL;DR

**Where global placement starts its cells is worth nothing to the finished
design.** Nine starting distributions — the core centre, the ODB position, two
corners, a uniform draw, two Gaussians, a deterministic low-discrepancy sweep
and a connectivity-aware seed — over 13 designs on 6 platforms, 3,120 samples.

1. **Nothing changes timing.** On the 10 designs where timing is measurable at
   all, **no arm resolves in either direction** on `min_period` or on TNS. The
   bound is a median 0.95 % of the achieved period.

2. **The effect is below the seed's own noise.** Choosing among the
   distributions moves `min_period` at **0.40×** the spread you get from
   redrawing `GPL_RANDOM_SEED` with one distribution. You cannot act on it:
   any single run draws from a wider distribution than the effect itself.

3. **But it does move the placement**, by about **1.4×** the seed's spread on
   HPWL and routed wirelength. So the start visibly changes what gets built and
   not how good it is. The seed converts placement disruption into timing
   disruption roughly 3.7× more efficiently than the start does — a small
   per-cell jitter scrambles relative positions along a path, a different start
   produces a globally different but internally coherent placement.

4. **The sentry is empty.** ORFS pins `-force_center_initial_place` flow-wide,
   added in January 2026 purely to preserve behaviour after OpenROAD changed
   gpl's default, with no measurement attached. Applying one centre policy at
   both global placement calls (`center`) **does not resolve** against that
   split on any endpoint. Whatever the flag is doing, it is not costing
   anything, and nobody needs to argue about it again.

5. **No distribution is worth switching to.** `gauss_wide`, `spread` and
   `uniform` are *worse* on HPWL; `corner_ll` and `gauss_tight` are outright
   inconsistent across designs. `anchored` — the connectivity-aware seed built
   specifically as a candidate fix — is inconsistent on HPWL and has a +4.77 %
   worst case on timing. **No upstream candidates are proposed.**

6. **Initial place does not converge on the large designs.** `asap7/jpeg`
   converged in **0 of 208** samples, `asap7/ethmac` in 10 of 208, both
   stopping at the 20-iteration cap. On small designs the residual is at 1e-7
   by iteration 1 and the loop runs only because of the `iter >= 5` floor. The
   break condition tests the BiCGSTAB residual of a single linearisation, not
   whether the bound-to-bound fixed point has settled. This is a property of
   the solver and survives any flow restructuring.

7. **One platform is far more start-sensitive than the rest, and it is not
   explained.** On `ihp-sg13g2`, scattering the start costs 20–25 % HPWL and
   starting in a corner *saves* 5–12 %, with non-overlapping ranges and no
   divergence reverts. Eleven other designs show nothing outside their own
   noise band. HPWL is measured before any repair, so this is not a censoring
   artefact. Reported, not explained.

8. **The obvious mechanism is wrong.** Non-convergence does not predict
   sensitivity: correlation between a design's capped-sample fraction and the
   spread of its arm means is **r = −0.176 over 13 designs**. The two designs
   that essentially never converge sit mid-pack; the two most sensitive
   converge on every sample.

### Why this bears on pinning macro placement

The question that started this was whether a replayed (pinned) macro placement
costs QoR. Under `-place_ios` (OpenROAD #11134, merged) the two global
placement calls collapse into one, and the single call starts from whatever the
floorplan left — exactly the pinned case. This study bounds that channel: **if
pinning costs QoR, it is not because global placement lost a good starting
position.**

## Method

* **A sample** is a place stage (tier 1) or place → cts → grt (tier 2) on a
  frozen floorplan. One `//:deps` tree per design; every sample runs inside it
  under its own `FLOW_VARIANT`, so all arms share the synthesis and floorplan
  bytes and differ only in the knob.
* **Arms** select a distribution at run time from one binary via
  `-initial_position_mode`. Passing no mode takes the shipped code path
  unchanged, which is the baseline.
* **Witness.** The patch logs the mode it ran. A sample whose log does not
  attest to the arm it was run as is **discarded**; the count is in the report.
  Tier 1 discarded **0 of 2704**.
* **Determinism**, checked not assumed: the same arm at the same seed twice
  gives a byte-identical `3_3_place_gp.odb`, and different arms do not.
* **Machine**: 32 threads, 122 GB. `NUM_CORES=4`, identical for every sample.
  Samples run several-wide, so wall times are recorded and never quoted;
  `gp_iterations` is the machine-independent runtime endpoint.
* **Seeds**: 16 per arm, 32 for the three whose start is itself a draw.
  Resolution is `2σ·√(2/k)`.
* **Concordance.** Nine arms on 13 designs is 117 comparisons; a 2σ
  per-comparison threshold produces about five confident-looking rows from
  nothing. An arm is a finding only when at least **4 of 13** designs resolve
  in the same direction — a threshold computed to hold the family-wise rate
  near 0.4 %, printed next to the rate it buys. Where both directions clear the
  bar the verdict is **inconsistent**, not the majority direction.

### Two conventions this study had to learn

* **Timing is normalised by the achieved period, never by the SDC clock.** The
  constraint is a number someone wrote down: on `sky130hd/gcd` the design lands
  at 2.37x its SDC clock, on `nangate45/gcd` 1.70x. Normalising by the
  constraint inflates each design differently and corrupts exactly the
  cross-design medians this study is built on.
* **A design that closes cannot measure timing.** Once WNS reaches zero
  `repair_timing` stops, so `min_period` records where the optimizer quit.
  `ihp-sg13g2/gcd` closes on 32/32 samples and `ihp-sg13g2/spi` on 32/32 --
  bottoming out at exactly 0.000. Those designs, plus `asap7/ibex` (1/32), are
  excluded from timing endpoints, and the report names them. In an earlier pass
  a censored design supplied the largest single timing effect in the study and
  reached the summary table as a finding. bazel-orfs PR #981 found the same
  censoring in ORFS's own `period_padding` rules.

### The noise floor, cross-checked

This campaign's measured HPWL spread is **0.65 %–3.09 % of the mean, median
1.55 %**. PR #977 measured 0.65 %–2.81 %, median 1.46 %, on a partly different
design set on the same host. The two agree, which is the only reason the
sizing above can be trusted.

## Results

Each arm's samples against the baseline ensemble, one panel per design. On eleven of thirteen designs every arm sits inside the baseline's own 2-sigma band; the two ihp-sg13g2 panels show the scattered starts 20-25% above it and the corner starts 5-12% below.

### Tier 1 — the place stage, 2704 samples, 9 arms, 13 designs

| arm | global-place HPWL (um) | global-place iterations |
| --- | --- | --- | --- | --- | --- | --- | --- |
| anchored | inconsistent 2/13, 2 opposed | better 7/13, 2 opposed | did not resolve 1/13 | did not resolve 1/13, 1 opposed | better 4/13, 2 opposed | inconsistent 2/13, 2 opposed | inconsistent 3/13, 2 opposed |
| center | inconsistent 3/13, 1 opposed | did not resolve 2/13, 1 opposed | - | - | - | - | - |
| corner_ll | inconsistent 6/13, 5 opposed | worse 6/13, 3 opposed | - | - | - | - | - |
| corner_ur | better 5/13, 3 opposed | inconsistent 6/13, 4 opposed | - | - | - | - | - |
| gauss_tight | inconsistent 3/13, 2 opposed | inconsistent 5/13, 5 opposed | - | - | - | - | - |
| gauss_wide | worse 5/13 | inconsistent 5/13, 4 opposed | - | - | - | - | - |
| odb | did not resolve 2/13 | better 6/13, 2 opposed | did not resolve 0/13 | did not resolve 0/13 | did not resolve 1/13 | did not resolve 0/13 | did not resolve 0/13 |
| spread | worse 6/13, 3 opposed | better 7/13, 3 opposed | better 4/13 | did not resolve 3/13 | worse 5/13, 3 opposed | worse 6/13, 2 opposed | inconsistent 5/13, 5 opposed |
| uniform | worse 5/13, 2 opposed | inconsistent 6/13, 4 opposed | - | - | - | - | - |

_Timing and routing columns are blank for the six arms tier 2 did not carry forward._

## Does initial place converge?

`doBicgstabPlace()` breaks when `residual <= 1e-5 && iter >= 5`, against an `-initial_place_max_iter` cap of 20.

| design | samples | iterations | final residual | converged | hit the cap |
| --- | ---: | ---: | ---: | ---: | ---: |
| asap7/aes | 208 | 5-6 | 9.65e-06 | 208 | 0 |
| asap7/ethmac | 208 | 17-20 | 6.21e-04 | 10 | 198 |
| asap7/gcd | 208 | 5-5 | 1.20e-07 | 208 | 0 |
| asap7/ibex | 208 | 5-11 | 9.90e-06 | 208 | 0 |
| asap7/jpeg | 208 | 16-20 | 1.32e-03 | 0 | 207 |
| asap7/riscv32i | 208 | 5-6 | 5.46e-06 | 208 | 0 |
| asap7/uart | 208 | 5-5 | 1.20e-07 | 208 | 0 |
| gf180/riscv32i | 208 | 5-20 | 1.12e-03 | 164 | 44 |
| ihp-sg13g2/gcd | 208 | 5-5 | 1.20e-07 | 208 | 0 |
| ihp-sg13g2/spi | 208 | 5-5 | 1.20e-07 | 208 | 0 |
| nangate45/gcd | 208 | 5-5 | 1.20e-07 | 208 | 0 |
| sky130hd/gcd | 208 | 5-5 | 1.20e-07 | 208 | 0 |
| sky130hs/gcd | 208 | 5-5 | 1.20e-07 | 208 | 0 |

## Where each global placement call starts from

GPL-0051's own counters, on the baseline arm.

| design | 3_1 odb | 3_1 core | 3_3 odb | 3_3 core |
| --- | ---: | ---: | ---: | ---: |
| asap7/aes | 1059 | 10027 | 0 | 10414 |
| asap7/ethmac | 0 | 57242 | 0 | 52745 |
| asap7/gcd | 43 | 261 | 0 | 314 |
| asap7/ibex | 0 | 18596 | 0 | 18839 |
| asap7/jpeg | 6038 | 45663 | 0 | 45708 |
| asap7/riscv32i | 8783 | 0 | 0 | 8913 |
| asap7/uart | 0 | 619 | 0 | 624 |
| gf180/riscv32i | 0 | 6214 | 0 | 6378 |
| ihp-sg13g2/gcd | 34 | 329 | 0 | 382 |
| ihp-sg13g2/spi | 0 | 137 | 0 | 159 |
| nangate45/gcd | 0 | 513 | 0 | 551 |
| sky130hd/gcd | 0 | 266 | 0 | 319 |
| sky130hs/gcd | 0 | 385 | 0 | 430 |

The same arms measured on min_period. Every panel's spread sits inside the baseline band, which is the null this study reports.

### Tier 2 — through global route, 416 samples, 4 arms, 13 designs

| arm | global-place HPWL (um) | global-place iterations | min_period = clk - WNS | setup TNS at grt | global-route wirelength | global-route usage % | global-route worst layer % |
| --- | --- | --- | --- | --- | --- | --- | --- |
| anchored | inconsistent 2/13, 2 opposed | better 7/13, 2 opposed | did not resolve 1/13 | did not resolve 1/13, 1 opposed | better 4/13, 2 opposed | inconsistent 2/13, 2 opposed | inconsistent 3/13, 2 opposed |
| odb | did not resolve 1/13 | better 6/13, 2 opposed | did not resolve 0/13 | did not resolve 0/13 | did not resolve 1/13 | did not resolve 0/13 | did not resolve 0/13 |
| spread | worse 5/13, 3 opposed | better 6/13, 3 opposed | better 4/13 | did not resolve 3/13 | worse 5/13, 3 opposed | worse 6/13, 2 opposed | inconsistent 5/13, 5 opposed |

## What did not move the needle

* **Every hypothesis I had about the mechanism.** I predicted that all cells
  starting on one point would be degenerate for a bound-to-bound net model, and
  that spreading them (`spread`, the deterministic low-discrepancy sweep) would
  therefore help. The data says the reverse: concentrated starts are fine and
  scattered starts are the ones that hurt.
* **"HPWL is not a proxy for timing."** I wrote this down mid-campaign off
  the `ihp-sg13g2/gcd` result above, and the very next design refuted it. Kept
  here because the shape of the error is the useful part: a striking number on
  one design reads as a general law until the concordance rule is applied to
  it, and the rule only protects you if it was written before the data arrived.
* **Convergence as the explanation.** The initial-place convergence probe was
  built to be the mechanism, and it is not one: r = −0.176 between a design's
  capped fraction and its start sensitivity. The convergence table below is
  still a true and, I think, interesting fact about gpl — the break condition
  tests the BiCGSTAB residual of one linearisation, which is at 1e-7 by
  iteration 1 on small designs, rather than whether the bound-to-bound fixed
  point has settled — but it does not predict QoR sensitivity and is not
  offered as if it did.
* **`shipped` being cheapest.** I predicted the inherited centre policy would
  at least win on runtime. `odb` and `anchored` both beat it.
* **`anchored`**, the connectivity-aware seed built specifically as a candidate
  fix — centroid of the fixed pins a cell sees through its own nets. Better on
  iterations (7/13) but inconsistent on HPWL, and it anchors only a minority of
  instances (87 of 314 on `asap7/gcd`). Measured, not carried.
* **`corner_ll` / `corner_ur`.** Interesting on one platform, incoherent across
  the fleet: `corner_ll` is inconsistent 6/13 with 5 designs opposed. A real
  effect whose sign depends on the design is not a policy.

## Limits, written before anyone asks

* **No macro design.** `asap7/tinyRocket` was in the set and fails synthesis on
  an unrelated AUTO_MEMORIES error; it was replaced by two macro-free designs.
  Everything here is measured on designs where `rtl_macro_placer` does not run,
  so it says nothing about the macro-placement hand-off that motivated the
  question.
* **13 designs, 6 platforms**, weighted toward small ones. The two designs that
  fail to converge in initial place are the two largest, so the regime that
  matters most is the one with the least coverage.
* **HPWL and iterations are proxies**, and tier 2 says they largely do not
  reach timing.
* **Ten designs measure timing, not thirteen.** Three are censored by closure,
  and two of those are the `ihp-sg13g2` pair that shows the largest HPWL
  sensitivity -- so the most interesting designs on one axis are unmeasurable
  on the other.
* **Nothing here is a runtime study.** Samples ran several-wide, so wall times
  are recorded and not quoted; `gp_iterations` carries the runtime claims. The
  one runtime number quoted -- the 5.5 s that withdrew the candidate -- is an
  order-of-magnitude argument, not a measurement.
* **`-place_ios` is not measured.** OpenROAD #11134 collapses the two global
  placement calls into one, which removes the stage that currently launders the
  start position. Every result here describes the flow as it is today.
* **The `ihp-sg13g2` sensitivity is reported, not explained.**

## Upstream candidates: none

Nothing is proposed, and nothing was filed anywhere.

An earlier draft of this study carried one candidate — letting `3_3_place_gp`
warm-start from `3_1_place_gp_skip_io` instead of discarding its converged
placement, which it does on all 13 designs. It is withdrawn for two independent
reasons, either of which is sufficient:

* **It is not meaningfully better.** Summed over 13 designs it saves 5.5 s of
  157.6 s of global placement — and global placement is a median 14 % of the
  place stage, so about **0.15 % of place → cts → grt**. Per design it is
  incoherent: `asap7/ibex` +5.8 % and `asap7/riscv32i` +8.0 % *slower*.
  "Better 6/13" was a statement about resolvability, not about size.
* **The stage is going away.** `-place_ios` collapses `gp -skip_io` →
  `place_pins` → `gp` into one pass, so `3_1_place_gp_skip_io` — and with it
  the placement there is to warm-start from — does not survive.

**Chesterton's Fence.** ORFS did not pin `-force_center_initial_place` out of
carelessness: OpenROAD changed the gpl default underneath it in #9353, and the
flow-wide flag was the minimal change that kept behaviour identical while the
dust settled. That was a reasonable call, and it is exactly why no number was
attached to it. What this study adds is the number, and the number is zero.

Resolvable difference against seeds per arm, one line per design, log x-axis. The knee is at 16 seeds; quadrupling k only halves the resolution.

## Reproducing

See `docs/studies/gpl-initial-position/README.md` for the full method, the
arms, the sizing arithmetic and the traps. The short version:

```sh
bazelisk run //:deps -- @orfs//flow/designs/asap7/gcd:gcd_place
bazelisk run //test/gpl_initial_position:campaign -- --help
bazelisk run //test/gpl_initial_position:report   -- --results <dir>
bazelisk test //test/gpl_initial_position:all     # no ORFS checkout needed
```

Raw samples are committed on this branch rather than pasted as a comment --
2,704 and 416 rows are past what a comment holds:
[`tier1_samples.csv`](https://raw.githubusercontent.com/The-OpenROAD-Project/bazel-orfs/7d7c6c02b1bb167bbfc6e4af05906debcebda49f/docs/studies/gpl-initial-position/tier1_samples.csv) and
[`tier2_samples.csv`](https://raw.githubusercontent.com/The-OpenROAD-Project/bazel-orfs/7d7c6c02b1bb167bbfc6e4af05906debcebda49f/docs/studies/gpl-initial-position/tier2_samples.csv), one row per `(design, arm, seed)`.
Every table above recomputes from them.


oharboe:
Closing as intended, reference only. Self-contained: the instrumentation patch, the harness, the figures and the raw data are all on this branch, and the write-up is in the PR body above. Nothing here is proposed for merge, and the branch is not deleted since the body's figures are pinned to a commit on it. No upstream repository was written to.

**No upstream candidates.** One was carried through most of the campaign — letting `3_3_place_gp` warm-start from `3_1_place_gp_skip_io` rather than discarding its converged placement — and is withdrawn twice over: it is worth ~0.15% of place → cts → grt, and `-place_ios` ([OpenROAD #11134](https://github.com/The-OpenROAD-Project/OpenROAD/pull/11134)) removes the stage it would improve.

**Two things a reader might want from this branch even though the finding is a null:**

1. `test/gpl_initial_position/stats.py` — the concordance rule. Nine arms on thirteen designs is 117 comparisons, and a 2σ per-comparison threshold produces about five confident-looking rows from nothing. The rule computes its own threshold from the design count and prints the family-wise rate next to it, returns `inconsistent` when both directions clear the bar rather than reporting the majority, and returns `underpowered` when the design count cannot support any verdict at all.

2. `test/gpl_initial_position/report.py` — the censoring screen. Once WNS reaches zero `repair_timing` stops, so `min_period` records where the optimizer quit. Three designs here are censored; in an earlier pass one of them supplied the largest single timing effect in the study and reached the summary table as a finding. This is the same censoring [#981](https://github.com/The-OpenROAD-Project/bazel-orfs/pull/981) found in ORFS's own `period_padding` rules.

**One bug found in a sibling study's harness**, reported rather than acted on: `elapsed_seconds` in `test/gpl_seed_qor/harvest.py` on `study/gpl-seed-qor` reads ORFS's `[h:]min:sec` as if the first field were hours, inflating any non-zero value 60-fold, and does not match the three-field form at all. [#977](https://github.com/The-OpenROAD-Project/bazel-orfs/pull/977) published no runtime column so its data is unaffected, but anyone reusing that harness for a runtime question would inherit it. Fixed on this branch with a test.

oharboe:
@maliberty FYI, no action required. I was curious as to whether starting positions improve QoR or running times and l was unable to reproduce it. TL;DR starting positions of stdcells is below the global placement random seed noise

PR #986, opened 2026-09-14
study: CoreMark/Joule for four RISC-V cores at one stated core+L1 boundary

oharboe:
CoreMark/Joule against CoreMark/MHz

*Figure 1. CoreMark/Joule against CoreMark/MHz, both axes logarithmic.
Blue: measured here on ASAP7 at global route, activity from one hot
CoreMark iteration, every point at the same core-plus-L1 boundary. Red:
a published GF 22 FDX series, drawn in its own colour because it is not
like-for-like.*

> **Work in progress.** Every measurement target here is
> `tags = ["manual"]`. The numbers are real and reproducible, and the
> parts that are modelled rather than measured are named as such in §5.
> This is the first version, not the final one.

CoreMark/MHz is reported for almost every open-source RISC-V core;
CoreMark/Joule almost never is, because the energy half needs a
hardened netlist, a switching-activity capture, and a power engine, and
each of the three has a way of producing a plausible number that is not
a measurement. We build the whole chain in a reproducible flow —
CoreMark ELF, RTL simulation, CRC gate, synthesis, global route,
gate-level simulation, SAIF over one hot iteration, `report_power` —
and screen at global route so a point costs minutes rather than hours.
We report four cores on ASAP7: SERV (0.0243 CoreMark/MHz), picorv32
(0.5531), ibex (2.4543) and VeeR EH1 (4.7979), spanning nearly two
hundred times in performance per clock.

**Every point is measured at the same stated boundary — the core and
its L1, and nothing beyond it — and the boundary is verified rather
than asserted.** Cores with no cache are hardened together with the
tightly-coupled memory they execute from; each wrapper counts every
transfer that leaves the hardened block, and for all four cores one hot
CoreMark iteration sends **zero**. Closing that boundary is itself the
study's largest result: it costs the cacheless cores between **2.8x and
4.2x** of their CoreMark/Joule, with CoreMark/MHz unchanged to every
digit, and it reverses the ordering — ibex reads 8.08x better than VeeR
EH1 when its memory is outside the measurement and 2.89x better when it
is inside. Any CoreMark/Joule quoted for a small core without stating
whether its memory was measured is uninterpretable at roughly an order
of magnitude.

The study's central methodological contribution is negative and
checkable. OpenSTA does not fail when a pin carries no annotated
switching activity: it estimates one, and the estimate is
indistinguishable from a measurement in the report. We therefore
enumerate every pin, classify every pin the SAIF did not reach, and
bound what the estimator could be worth by sweeping the default
activity across its entire range. For the three cacheless cores the
SAIF annotates **100 % of pins** (27,597 / 52,280 / 82,396), **zero**
are unannotated, and for all four cores the SAIF-driven total is
bit-identical at ten significant figures across the whole sweep, while
the same sweep moves the vectorless total by 23--179 %. The energy
numbers are therefore vector-driven in the strong sense: OpenSTA's
probabilistic activity model contributes nothing to them.

The shape the points make is reported with its own diagnosis, and the
diagnosis was tested. An earlier draft found the cacheless cores on a
straight line in log--log axes and predicted, falsifiably, that the
boundary was the cause. Hardening the memories confirmed most of it —
extrapolating the cacheless trend to VeeR's performance overpredicted
its efficiency by **15.2x** before and **5.65x** after, closing 63 % of
the gap — and refuted the rest: the line is still there, and tighter.
The reason is now the platform's memory model rather than the boundary,
and we report that inversion rather than the confirmation alone, because
it is what the data supports: **CoreMark/Joule is not yet a
discriminating axis among cores of this class.**

We state, rather than imply, what the numbers do not yet cover. Every
memory is a FakeRAM abstract, and FakeRAM's ASAP7 views charge **the
same switching energy and the same leakage for every shape** — so the
58--71 % of each point's power that is memory is a term that does not
know how large the memory is. Halving every memory in this study would
move no number in it. That is deliberate: one model applied uniformly is
something a comparison survives, and two models is not — but it is the
single largest thing between this work and an energy measurement, and it
is being fixed in the generator rather than worked around here. ibex is
still measured with `ICache=0`. The simulation is zero-delay and so
carries no glitch power, the parasitics are estimated rather than
extracted, the corner is ASAP7's best case, and the frequency is an SDC
target rather than an achieved maximum. Each is quantified or bounded in
§5.

---

The full paper is [`test/coremark_joule/README.md`](https://github.com/The-OpenROAD-Project/bazel-orfs/blob/study/coremark-joule/test/coremark_joule/README.md);
this body is its abstract.

**Draft on purpose.** This branch is a first version to be whittled down:
the bazel-orfs and ORFS fixes it carries are being split out as
single-concern PRs, and the modelling gaps named in the abstract are the
work after that.


oharboe:
@maliberty Just FYI: I'm working on a test/example/study of a range of RISC-V CPUs where I plot work/time on the x axis and compute/energy on the y axis.

It covers 0.02(single bit RISC-V) to 10 CoreMark/MHz(in the range of x86/arm), so 4 orders of magnitude in performance.

This will of course exercise a very large number of fiddlybits in OpenROAD/OpenSTA/bazel-orfs...

oharboe:
For anyone following this thread: continued in #1005, which is this work rebuilt on current `main` as single-concern commits, CI-green. The `study/coremark-joule` branch stays on the remote, so the commit-by-commit history of how the measurement was arrived at remains readable.

Further work continues as one PR per concern on top of #1005 — first up is deriving each core's achieved frequency rather than scoring it at an SDC target.

PR #988, opened 2026-09-15
study: XiangShan V3 — 8.29 CoreMark/MHz; 34 blocks hardened as macros, the parent through CTS

oharboe:
XiangShan V3 (Kunminghu) as rung 10 of the CoreMark-per-Joule study. Rebased onto today's `main`; the infrastructure the study needed has been split out and merged (list below), so this PR is now the study itself: the XiangShan design, its Chisel patches, the 34-block hierarchical configuration, and the README sections.

## Status, 2026-09-18

```
one iteration  120,582 cycles   ->  8.2931 CoreMark/MHz      (measured, CRCs correct)
```

The x-axis is measured. The y-axis has reached the parent's clock tree: every block is placed and abstracted, the parent is synthesised, floorplanned, placed and CTS'd, and its global route is the open problem, quantified below.

| parent (1.5 M instances, 44 macros, 3.6 mm die, 33 % util) | take 13 |
|---|---|
| synthesis, 34 blocks in parallel, slowest block | 9 min |
| block abstract (`write_timing_model`), slowest | 21 min (LsqWrapper) |
| floorplan incl. macro placement | 16 min, 0 row fragments under 20 µm |
| global placement | 17 min |
| repair_design | 10 min |
| detailed placement, diamond legaliser | 149 s, max displacement 388 µm |
| CTS | 685 s, 22 790 clock buffers |
| post-CTS legalisation, negotiation legaliser | 1050 s, 400 iterations |
| peak memory, CTS stage | 59 GB |
| global route | stopped after 7 h in maze iteration 1 of 30, 105 GB working set |

Take 12's parent had failed detailed placement on 4 755 wire buffers stranded in 22 000 row fragments 5 to 20 µm wide between abutting macros; 54 µm channels (`ANNEAL_BLOCK_GAP_UM`), 33 % utilisation and the diamond legaliser with a macro-wide window fixed that, and the ODB forensics that found it became `odb-debug` (#1048).

## Global route is the pathology, and it is the die

A 15-second perf sample of the router: 95 % of cycles in FastRoute's monotonic pattern routing and its edge-usage reads; later, 94 % in the sequential maze router, of which 69 % is one `std::ranges::find` over the Dijkstra heap (`maze.cpp:1286`), a linear scan of the frontier for every decrease-key, still on upstream master. OpenROAD sizes a GCell as 15 track pitches of metals 2 to 4 (hard-coded in `odb::dbBlock::getGCellTileSize`): 0.54 µm on asap7, so this die is a 6 700 by 6 700 grid, 45 M cells per layer, and every phase and most of the 48 GB the router adds scale with it. `pin_access` before routing took 923 s.

The plan, in order, with a small-scale reproducer first: a wirebound variant with IO anchors on a swept die (`DIE_AREA` 0.5 to 3.6 mm), a bench script that records FastRoute's phase timers, peak RSS, overflow and wirelength, a matrix driver, and a bring-your-own OpenROAD binary. Then an unattended matrix on the parent's CTS ODB over the knobs that already exist (`-congestion_iterations 5 -allow_congestion`, `-snapshot_batched_width 48`, CUGR, fewer routing layers, huge pages), and only then the code patches the numbers justify: a heap with `(cost, cell)` pairs and a position index, cache blocking in `routeMonotonic`, late allocation of the 3D arrays, a GCell-pitch knob. One concern per patch, measured on the reproducer, then on XiangShan.

## Split out and merged

#1029 `tools/structured_gen`; #1039 patches 0072–0074 (AUTO_MEMORIES by module, FakeRAM declared pins, SKIP_EXTRACT_FA); #1041 patches 0075–0076 (every ORFS variable names its stages, DETAIL_PLACEMENT_ARGS in cts and grt); #1043/#1044 design DSL forwards user_stages, block_abstract_stage, quick_pins, canon_blackbox_macros; #1045 bazelignore; #1046 simulator from srcs; #1047 macro_anneal; #1048 odb-debug; #1049/#1050 agent skills.

Left in this PR from the infrastructure side: `orfs_design` lets BLOCKS flows read the parent's user variables (one commit, a breakout candidate).

## Carried, not upstreamed

Seven XiangShan Chisel patches under `test/coremark_joule/xiangshan/patches/`: the vl CSR read without the difftest RAT buffer (upstream as OpenXiangShan/XiangShan#6589, draft), Ftq queues and the rename buffer as generated register files, Rob and store-queue entries as one module each, store-to-load forwarding per load pipeline, the Rob cell's packed input bus. Each header says what it fixes and how it retires.

## What a reader should know

- **Turnaround settings are on** in the parent's `config.mk` and the blocks' `block.mk`, each flagged as such; the measured run flips them.
- **Frequency is the SDC period**, 1200 ps, not an achieved maximum.
- **The macro power model is FakeRAM's**, a constant per macro, the same model the platform's own views were generated with.
- **Blocks are abstracted at placement** (`block_abstract_stage = "place"`), opaque at the top, every metal layer.


oharboe:
## What XiangShan specifically is teaching this flow, and the tool it points at

A note for the record, separate from the PR body, on where the pain concentrated once XSCore went hierarchical and what the idiomatic fix looks like. Nothing below is started; it is the plan.

### Where the pain is

Not in synthesis, not in global placement, not in the power grid. Every one of those had a setting or a known idiom: explicit kept lists per block, no in-loop repair or routability inflation, the annealer where RTL-MP fails at a few hundred macros, ORFS's BLOCKS grid and a 2 um margin at the parent. The pain is one structure, and it is the same in every region:

| register file | shape | ports | outcome as flops |
|---|---|---|---|
| FpRegFilePart0..3 | 4 x 256 x 16 | 14R 7W | fp region block failed legalisation twice (4 h 10 min, 1 h 17 min), 17 k misaligned cells, all under the fp data path |
| IntRegFile | 224 x 64 | 12R 9W | the parent's legaliser on the same curve: 5.5 M violations to 17 k in 400 iterations and 3 h, then ~1 k per 50 |
| VfRegFile | 128 x 128 | 14R 7W | converged inside the vector block, 38 iterations |

A many-port register file synthesised to flops is a flop per bit plus a 14-way read mux tree per bit and a 7-way write demux, and the legaliser cannot spread its local density whatever the average utilisation: 40 %, 28 % with two sites of padding and a 2000-site search window, same residue. A core competing in XiangShan's class does not synthesise these at all; they come from a register-file compiler as multi-port bitcells, and its TLBs and load/store-queue address matches from CAM macros. They are macros in the design and absent from its gate count, which also bears on the literature comparison in README §4.9: some of XiangShan's 7.4x over the C910's gates is flop register files measured against compiled ones.

The workaround in this PR is to make the register files blocks of their own, at low utilisation, so that legalisation happens where it is cheap. That is a macro flow doing a compiler's job with a placer; it closes the loop and it documents the "before".

### The tool: `tools/structured_gen`, a `cc_binary` in bazel-orfs

The idiomatic fix is a generator, like OpenROAD's flop-based RAM generator but multi-port and living here, since upstream is not where the shape should churn. It links OpenROAD's libraries (odb, utl, sta) rather than emitting text: it reads the platform LEF and liberty with OpenROAD's own readers, so site, tracks, cell footprints and pin shapes are the real ones, and it writes a placed, legal ODB plus a structural Verilog netlist for simulation.

- **Register file**, bit-sliced: word rows by bit columns; a tile per (word, bit) of one flop, a W-way write mux and R AND2 gates; row headers with the write decode and R one-hot read wordlines; column footers with the R OR trees that are the bitlines. Legal by construction; read path of seven or eight gate levels.
- **CAM**: key flop plus P XNORs per tile, per-entry AND tree in the row footer driving the match line, priority encoder at the edge on request. First targets once they have a module boundary: the 48-entry L1 TLBs and the load/store queue address matches.
- **Detection, post synthesis**, in the slot AUTO_MEMORIES already has. Tier 1: an inferred `$mem_v2` with more than one port is classified `regfile` and routed to the generator instead of refused. Tier 2: a flop-vector register file by module shape, registers of one width dominating, enables from address compares, mux-tree reads, with port roles classified by function and names only as a tie-break. CAMs explicit-mode first.
- **Loud failure is the contract.** Every candidate ends generated, refused with a one-sentence reason, or ignored by listing, in a `structured.json` beside `memories.json`. `STRUCTURED_BLOCKS = IntRegFile:regfile ...` must resolve and classify completely or the build fails there; a gate like `check_memories_applied` refuses a large register file synthesising as flops.
- **Bazel's analysis-time graph**: the detection is pinned into a `<design>.structured` file by a run target, block targets are generated from it, and a check action fails on drift, the same shape as the `.memories` override.
- **Flow**: an `orfs_generated_block()` rule presents the ODB as a placed stage, so CTS, route and abstract run unchanged and the parent takes the LEF and lib through `macros`. Simulation uses the structural netlist, so the SAIF annotates inside the macro.

Order and rough cost: register-file generator and explicit mode, three days; stage rule and pinning, one; tier-1 detection, one; tier-2, two; CAM and its XiangShan boundary patches, two. Explicit mode alone unblocks every region above.

oharboe:
@maliberty XiangShan WIP

PR #989, opened 2026-09-15
AUTO_MEMORIES: make it work with parallel synthesis

oharboe:
`AUTO_MEMORIES` generates `.lib`/`.lef` views for the memories it converts, and every later stage globs them back out of the results dir. The **serial** synthesis path stages those views into the actions that read them; the **parallel** path never did, in three places.

The parallel path is taken whenever `SYNTH_HIERARCHICAL=1`, so **`AUTO_MEMORIES` and hierarchical synthesis could not be used together at all.**

### The three gaps

1. **Partition actions did not stage the artifacts.** make's chain is

   ```
   yosys-dependencies: memories.json
   memories.json: memories_inferred.json ...
   memories_inferred.json: $(VERILOG_FILES) extract_memories.tcl
   ```

   so a partition sandbox without `memories_inferred.json` re-runs detection there — against a design already reduced to canonicalized RTLIL, with no Verilog left to read. It fails as ``Module `<top>\' not found!``.

2. **Staging is necessary but not sufficient.** A sandbox gives inputs mtimes make reads as out of date, so `yosys-dependencies` would remake `memories.json` — a read-only input there. `--old-file` is the "up to date whatever the timestamp says" switch, now passed for each staged artifact.

3. **The ODB action did not stage them either**, so `synth_odb.tcl` — which reads them via `load.tcl` and `read_liberty.tcl` — linked the blackboxed memory modules against nothing:

   ```
   [ERROR ORD-2013] instance <inst> LEF master <memory> not found
   ```

   The serial ODB action has staged them all along, with a comment naming this exact error.

### Test

`auto_memories_outputs_test` gains a `consumed` attribute.

The file already asserted the views are *declared*, which is necessary because only declared outputs survive a sandbox. But declaring an artifact that nothing consumes leaves it discarded at the boundary — and by the time OpenROAD looks for it, that is indistinguishable from never having been generated. That is the gap this bug lived in.

Coverage is partial and the docstring says so: `asap7/tinyRocket` is the only design here with `AUTO_MEMORIES=1`, and it does not set `SYNTH_HIERARCHICAL`, so it exercises the serial path — the one that was always correct. Catching this particular regression in an `analysistest` needs a design that sets both; adding a minimal one is worth doing separately.

### How it was found

A design with `AUTO_MEMORIES=1` and `SYNTH_HIERARCHICAL=1` that hardens a program memory with each core. It failed at ORD-2013 with correct views sitting in the results dir, generated and then discarded.


oharboe:
Force-pushed: the `consumed` assertion I first pushed was too weak and I want that on the record rather than quietly swapped.

It asserted that *some* action takes the generated views as an input. The partition actions do — that is the first of the three gaps this PR fixes — so the assertion passed with the ODB action still unfed. I confirmed that empirically: removing only the ODB staging left the test green.

It is now `consumed_by`, which names the reader:

```starlark
consumed_by = {
    "1_synth.odb": ["memories.json", "memories"],
},
```

`synth_odb.tcl` globs the views out of the results dir via `load.tcl` and `read_liberty.tcl`, so the action producing the ODB is the one that has to be handed them. With only the ODB staging removed the test now fails with:

```
the action producing 1_synth.odb in //test/auto_memories:am_top_synth does not
take memories.json as an input. Only declared outputs survive a sandbox, and only
staged inputs reach the action that reads them -- a stage that globs this out of
the results dir will find nothing.
```

Note the target in that message: it needs a design with `AUTO_MEMORIES=1` **and** `SYNTH_HIERARCHICAL=1`, and `asap7/tinyRocket` is not one. The assertion here still guards the serial path correctly, but the fixture that makes it bite on the parallel path is a separate PR stacked on this one.

PR #990, opened 2026-09-15
FakeRAM ASAP7: pick a column mux factor instead of hardcoding 1

oharboe:
The ASAP7 FakeRAM backend sets `"column_mux_factor": 1` in its process config and `"banks": 1` per memory, so a generated array is `rows` cells tall and `bits` cells wide with nothing folding it. Survivable for the shapes ASAP7 happens to ship, degenerate for anything deeper:

| shape | generated | aspect |
|---|---|---|
| 2048 × 32 | 4.18 × 663.60 µm | 159:1 |
| 8192 × 32 | 4.18 × 2654.40 µm | **635:1** |

A 2.65 mm tall macro cannot be placed in a die a fifth of a millimetre across, so `AUTO_MEMORIES` cannot produce a usable abstract for a memory of any size — which is the one thing it exists to do.

### The platform's own views are not built this way

Height scales as `rows/column_mux_factor` and width as `bits×column_mux_factor`. Reading the shipped LEFs back through that relation recovers the factor each was generated with:

| shipped | size | implied factor |
|---|---|---|
| `fakeram7_2048x39` | 20.33 × 166.60 µm | **4** |
| `fakeram7_256x32` | 8.36 × 42.00 µm | **2** |
| `fakeram7_256x256` | 33.25 × 84.00 µm | **1** |

So the factor is per shape, and the backend's constant 1 is what is wrong rather than the value it happens to hold.

### The fix

Pick the power of two that brings `rows/factor` closest to `bits×factor` — squaring the array. That reproduces the shipped choices for `256x256`, `64x21` and `128x64` exactly, and `256x32` and `2048x39` within one step.

The pathological cases come back in hand:

| shape | before | after | aspect |
|---|---|---|---|
| 8192 × 32 | 4.18 × 2654.40 | 66.50 × 166.60 | 2.5 |
| 2048 × 32 | 4.18 × 663.60 | 33.25 × 84.00 | 2.5 |

That second result is **byte-identical in geometry to the shipped `fakeram7_256x256`**, which is the cross-check that the folding now matches the generator that produced the platform's own views.

Area is unchanged by construction: column muxing moves cells between the two dimensions, it does not add or remove any.

### Why a separate patch from 0063

`0063` is ORFS's own backend patch, re-pathed onto the vendored copy rather than rewritten, so it stays comparable to the file ORFS ships. This is a behavioural change and does not belong inside it.


oharboe:
CI failure here is real, and it is not in this patch — it is in the checker that vets it.

`//test/bump:bump_test` fails with the conflict attributed to **0063**, not 0067:

```
patches/0063-orfs-fakeram-asap7-backend.patch: creates
tools/FakeRAM2.0/orfs_asap7/generate.py, which already exists
```

`test/bump/bump_test.py` synthesizes its fake ORFS tree from the patches themselves — a path a patch *modifies* gets a file, a path a patch *creates* is left absent. 0067 modifies `generate.py`, which 0063 creates, so 0063 started seeing it as already present.

That looked like a fixture quirk, so I checked `check_orfs_patches` against a genuinely pristine tree. Same blind spot in the real path:

```
patches/0067-orfs-fakeram-asap7-column-mux.patch:
  tools/FakeRAM2.0/orfs_asap7/generate.py does not exist
```

`repository_ctx.patch` applies `ORFS_PATCHES` in list order, so 0067 editing a file 0063 created is legal and works — the checker just models every patch as applying to the pristine tree. Landing 0067 as-is would make `//:bump` refuse commits that apply fine, recoverable only with `--ignore`, which silences the whole check.

Fixed in #999. **This PR is blocked on it** and should go green on a rebase once that lands.

I considered folding 0067 into 0063 instead, which would sidestep the checker entirely. Against it: 0063 is ORFS's own patch re-pathed rather than rewritten, so it stays diffable against what ORFS ships, and that property is worth more than the convenience. The checker being order-blind is also a bug on its own terms — it constrains how any patch series here can be layered.

oharboe:
Correction to my previous comment: the blocker is **#1000**, not #999.

Another session had already fixed the same bug, more completely — carrying modifications forward with a real `apply_hunks`, not just creations — so I closed #999 in its favour.

Verified: with `0067` registered on top of #1000, `//test/bump:bump_test` passes. This PR should go green on a rebase once #1000 lands.

PR #999, opened 2026-09-15
bump: check ORFS patches in the order bazel applies them

oharboe:
`repository_ctx.patch` walks `ORFS_PATCHES` **in list order**, so a patch is entitled to modify a file an earlier patch in the same list created. That is what carrying a vendored tool looks like: `0063` creates the FakeRAM ASAP7 backend — ORFS ships the patch and nothing applies it — and a later patch changing how that backend folds an array has to edit the file `0063` wrote.

`check_orfs_patches` read every patch against the **pristine** ORFS tree, so it rejected such a series:

```
patches/0067-orfs-fakeram-asap7-column-mux.patch:
  tools/FakeRAM2.0/orfs_asap7/generate.py does not exist
```

on a file that is there by the time the patch runs. `//:bump` then refuses a commit that would in fact apply, and the only way past it is `--ignore`, which silences the whole check rather than the one wrong verdict.

### What changed

Creation is carried forward: a file a patch creates is visible to the patches after it, **with the content the patch wrote**. That content is exactly the added lines of a creation hunk, so it needs no patch engine — and it means a later hunk is checked against what was actually created rather than merely against the file existing.

Modification is *not* carried forward, and the docstring says so. Two patches editing the same file are each checked against the original, which is correct whenever their hunks do not overlap. `ORFS_PATCHES` has that case today — `0060` and `0061` both touch `extract_memories.tcl` — and it passes. Doing better needs a patch engine, and the failure worth fixing was the create-then-modify one.

### Tests

Three, and two fail against the old behaviour with exactly the wrong verdict:

- a patch modifying what an earlier patch created → was `does not exist`, should be clean
- the same, with a hunk that does *not* match the created content → was `does not exist`, should be `hunk ... does not match`
- and a created file the tree already has is still caught, so carrying creation forward does not mask the real conflict

### How it was found

CI on #990, which carries `0067`. The failure surfaced as `//test/bump:bump_test` reporting the conflict against **`0063`** rather than `0067` — `test/bump/bump_test.py` synthesizes its fake ORFS tree from the patches themselves, so a path `0067` modifies became a path that "already exists" for `0063`. That made it look like a fixture quirk; checking `check_orfs_patches` against a genuinely pristine tree showed the same blind spot in the real path.

**#990 is blocked on this** and goes green once it lands.


oharboe:
Closing in favour of #1000, which fixes the same bug more completely and landed from another session in parallel.

Both make `check_orfs_patches` follow the order `repository_ctx.patch` actually applies the list. Mine carried **creation** forward only, and its docstring stated the limitation: two patches editing the same file are each checked against the original, which happens to be correct for ORFS_PATCHES today because 0060 and 0061 do not overlap. I judged a real patch engine to be more than the failure warranted.

#1000 writes that engine — post-images out of `parse_unified_diff` and an `apply_hunks` that carries modifications forward too. That is strictly better: it removes the limitation rather than documenting it, and it fixes `test/bump/bump_test.py`'s synthesized fake tree as well.

Verified it unblocks #990: with `0067` registered on top of #1000, `//test/bump:bump_test` passes.

Nothing here is worth salvaging into it. My three tests overlap what #1000 already asserts.

PR #1003, opened 2026-09-16
libxcb: drop the implicit __init__.py DEBUG warning

oharboe:
rules_python prints a DEBUG warning on every build:

```
WARNING: Target @@libxcb+//:c_client is using implicit __init__.py creation.
  This diabolic behavior is deprecated and will be disabled by default in a
  future release.
```

`@@libxcb+//:c_client` is a `py_binary` in libxcb's BUILD.bazel, which is a
BCR overlay file — so the fix is a patch on that overlay plus a
`single_version_override`, setting `legacy_create_init = 0` on the target.

It costs nothing here: the only package `c_client` imports, `xcbgen`, ships
its own `__init__.py`, and that file is inside `@xcb-proto//:py_xcbgen`'s
`glob(["xcbgen/*.py"])`. Nothing was relying on the implicit creation.

Verified by building `@@libxcb+//:c_client_xproto` — the genrule that actually
runs `c_client` to generate `xproto.c`/`xproto.h`. It succeeds and the DEBUG
line is gone.

Not upstreamed; retires on a BCR revision of libxcb that carries the attribute.


oharboe:
This PR's CI failure is not this PR's fault.

`//:guard_tool_test` fails on `main` too — I reproduced it locally on current `main` with a cold cache:

```
AssertionError: Cannot exec()
  '.../guard_tool_test.runfiles/_main/.claude/hooks/guard_tool_test.py': file not found.
```

The cause is `.bazelignore` listing `.claude`, while the root `BUILD` declares that test with `.claude/hooks/` sources — so bazel never stages them. This branch is based on `aade8a3`, the commit that introduced the ignore, which is why it is the first PR to show it: `main`'s own runs have been reporting cached passes for that target (the run after it executed 51 of 190 tests).

#1008 fixes it by scoping the ignore to `.claude/worktrees`, which is what the comment beside it says it is for.

The change here is confined to libxcb's own `c_client` `py_binary` and cannot reach `guard_tool_test`. Once #1008 lands, a rebase should be all this needs.

PR #1007, opened 2026-09-16
coremark-joule: the paper leads with what it is and what is wrong with it

oharboe:
The paper's README now opens with the title as a claim and the three qualifiers that make it true, then the limitations in the order they move the numbers. Nothing here re-runs a flow; `results.json` is unchanged except for the dated uniqueness record.

**What changed**

- **Opening.** Title, dateline (September 2026), the three least-restrictive qualifiers under which this is the only study of its kind, what enters when each is dropped, the seven limitations ranked, one implausible cross-check flagged, and the CoreMark-only scope stated as a decision.
- **Numbers rendered, not typed.** `scripts/readme_numbers.py` renders Table 1 and every ratio the prose quotes from `results.json`; `readme_numbers_test` (not manual, milliseconds) fails on disagreement. It caught the palimpsest: boundary cost "2.8–4.2x" vs "5.7–9.8x", ibex over VeeR as 8.08x/2.89x/1.41x, §4.7 clock and macro figures from a superseded run, "a factor of 92" that is 80.
- **§4.5** gains four near-miss studies (Schiavone PATMOS'17, Gallmann CARRV'21, Djupdal arXiv 2025, Elsadek & Tawfik 2021) and a dated ladder of which qualifier excludes which. Two further qualifiers that would each suffice alone are noted: a sub-10 nm node, and spanning both sides of the out-of-order line once XiangShan lands.
- **§4.8** replaces the rack-level model with cross-checks. Performance halves agree with Gallmann to 4 %; the energy halves do not scale as node scaling predicts (ibex core-only 3.60 µJ/iteration at 7 nm vs their 3.40 µJ at 65 nm), reported as unexplained with the measurement that would settle it. Ordering against Djupdal's subthreshold study agrees; ratios differ in the direction each regime predicts.
- **§5.13** states no point has an error bar and what would measure one. **§5.9** states the exact compiler flags. Draft-history narration ("an earlier draft…") is removed throughout.
- **§8.5** names `tools/memory_macro_scaler` as the fix for the shape-blind memory model. **§8.6** describes replicating a publisher's compiler output to verify a published CoreMark/MHz when that compiler is unavailable.
- **§7** keeps VeeR's results; its wiring notes move to `designs/asap7/veer/README.md`.

**Decided without a ruling, stated here**

- Figure 1 keeps the GF22 series in its own colour.
- The functional-unit attribution stays as a waived method section.
- `study/xiangshan` also edits this README and will need to rebase onto this.

**Checks**

```
bazelisk test //test/coremark_joule/scripts/...   # 11 tests pass, readme_numbers_test included
bazelisk run //:fix_lint
```


oharboe:
Rebased onto `main` after the clock-period commit landed. Every core's frequency is now derived rather than chosen, so the paper's third limitation changed from "every frequency is a target somebody chose" to "one derived pass on one floorplan, worth at most 3.5 % on the energy axis, second pass not yet run", and it moved down the ranked list below corner, glitch and parasitics. Table 1 and every quoted ratio re-rendered from the new `results.json`; the number test caught one hand-typed value on `main` (picorv32's boundary factor is 3.22x, not 3.23x). §5.5 and §8.3 are taken from `main` as written; §4.6 keeps the ratios-only stance and adds the derived-period column.

PR #1014, opened 2026-09-16
coremark-joule: time the SAIF by the period the design is built at

oharboe:
A SAIF records real time, and OpenSTA reads it as **transitions divided
by duration**. So the period the simulator times a capture with has to be
the period the design is built at.

It wasn't. The simulator's period was a literal in `sim/BUILD.bazel`,
next to a docstring in `run.bzl` saying it "**must equal the period in
the design's SDC**" and that "nothing downstream can detect the mistake".
#1006 re-derived every SDC period and left the literals alone:

| core | SAIF time base | SDC period | toggle rates |
|---|---|---|---|
| SERV | 700 ps/cycle | 438 ps | **1.60× low** |
| picorv32 | 1000 ps/cycle | 467 ps | **2.14× low** |
| ibex | 1200 ps/cycle | 1282 ps | 1.07× high |
| VeeR | 1600 ps/cycle | 1591 ps | 1.01× high |

This is my regression, introduced in #1006.

## What it cost, and why that is smaller than the ratio

Only **combinational switching** rides on data-pin densities. A
sequential cell's internal power is dominated by its clock pin, and a
macro's by its own clock pin, and OpenSTA takes both from the SDC at
`2/period` rather than from the SAIF. So the error reached ~4 % of
picorv32's total and moved its reported power by 2.9 %.

That is also **why it was invisible**: the macro column is 59–71 % of
every point and scaled correctly with the new periods, so the
re-measurement looked right while a smaller term underneath it stayed
frozen at the old clock.

| core | power | CoreMark/Joule |
|---|---|---|
| SERV | 43.40 → **44.30 mW** | 1,277 → **1,251** |
| picorv32 | 44.40 → **45.70 mW** | 26,676 → **25,918** |
| ibex | 19.50 → **19.00 mW** | 98,176 → **100,760** |
| VeeR | 86.50 → **86.90 mW** | 34,863 → **34,702** |

## The fix

`scripts/sdc_period.py` is now the only reader of `set clk_period`, and
the SAIF's time base, the reported frequency and the period tuner all
call it. Two copies of this one fact have now drifted — the frequency
literal fixed in #1006, and this one — so the answer is deleting the
copy, not adding a fourth warning comment beside it.

## Corroboration

The fitted figures move back to where they were *before* the periods were
derived: slope 0.954, R² 0.9998, `f/P` spread 1.26× across the three
cacheless cores. The frequency correction and this defect had been
pulling them in opposite directions. The power spread across those three
is genuinely wider now, 1.36× → 2.41×, because SERV and picorv32 run at
2283 and 2141 MHz against ibex's 780.

Every point re-measured end to end with a correct time base; the
annotation audit and activity sweep re-run on the new SAIFs. SAIF-arm
spread is still **0.0000 %** on all four, and §5.3b records the defect in
the paper rather than only the fix.

```sh
bazelisk test //test/coremark_joule/...   # 15 tests pass
```


oharboe:
Closing as a duplicate of #1010, which found the same defect three hours earlier and fixes it more narrowly (and with a `saif_window_test` case this one lacked).

Not wasted, though: I had written my fix independently before finding #1010, and re-running the entire measurement against #1010's implementation reproduced my numbers byte-for-byte — `pin_results: results.json unchanged`. Two diagnoses, two implementations, identical power on all four cores. That is a stronger check on the correction than either PR could give alone.

The part of this PR that was additive — the re-measured `results.json`, the paper's corrected tables and the plot — is now #1015, stacked on #1010.

PR #1016, opened 2026-09-16
memory_macro_scaler: LEF pins at ASAP7's wire width, on the pitch

oharboe:
Found by the CoreMark/Joule re-baseline's floorplan derivation: with scaler-generated macros, detailed routing left 558 to 717 DRC violations on every candidate floorplan of the picorv32 and ibex tiles, all of type `Rect Only` on M4, all at the macro's pin columns.

**Cause.** Signal pins were emitted as 0.048 µm squares on M4, one pitch. ASAP7's M4 is `WIDTH 0.024` with a width table of {0.024, 0.12, 0.216, ...}; 0.048 is not a legal width, so every wire that lands on a pin makes a non-rectangular shape. The platform's own `fakeram7_*.lef` uses 0.024 squares on M4 at pitch multiples and routes clean. The M5 pitch constant was also 0.068 where the tech LEF says 0.048, so clock pins were off-track.

**Fix.** Pins are the layer's wire width (0.024) and stay on the pitch; M5 pitch corrected. The test now asserts every signal pin is a layer-width square on its grid.

Checks: `bazelisk test //tools/memory_macro_scaler/...` passes; `fix_lint` run.


oharboe:
Amended: the M5 clock pin is now centred on its track. M5 is vertical with OFFSET 0, so a 0.024 pin whose left edge sits on the pitch multiple covers only half the wire, and detailed routing on picorv32 failed with DRT-0255 on the clock net. M4 pins keep the platform's fakeram7 convention (bottom edge on the pitch multiple; OFFSET 0.003 puts the wire centre inside the pin), which routes clean. Test updated accordingly; a picorv32 detailed route on this LEF is running to confirm.

oharboe:
Amended again, and this is the version that routes: clock pins move to M4 on the left edge with the inputs, the platform's fakeram7 convention. Centring the M5 pin on its track was not enough -- a 24 nm pin on a vertical layer at the top edge has no legal access unless its absolute x lands on an M5 track, and macro placement snaps to sites, not to the M5 pitch, so detailed routing still failed with DRT-0255 on the clock net. On M4 the pin sits on a pitch multiple relative to the macro exactly as fakeram7's do, and those route clean. Test updated; a picorv32 detailed route on this LEF runs next.

oharboe:
Confirmed on picorv32 through detailed route with this LEF: **0 violations** (the router converged 3254 → 341 → 182 → 5 → 0), against 600 left standing with the 0.048 µm pins and a DRT-0255 hard failure with the centred M5 clock pin.

oharboe:
Amended: the VDD/VSS stripes ran from x=0 across the whole macro, through both pin columns, and one signal pin per macro shape landed on a stripe's row, so the router had no access point for it (DRT-0073 on three pins of the study's VeeR tile). Stripes now start one pitch in from each edge, as fakeram7's do, and the pin banks step over the stripe rows with a pitch of clearance either side. The test asserts both. picorv32 already routed DRC-clean before this; VeeR grt is being re-run.

oharboe:
With the amended LEF, the study's VeeR tile (three sram_* shapes) now routes through detailed routing with 0 DRC violations; picorv32 already did. The DRT-0073 access-point failures are gone.

PR #1020, opened 2026-09-16
study: repair_timing's information is better, its ranking is not, and on this vehicle the repair itself is worth 1 ps and 2.7% of area

oharboe:
**This is a documentation PR: it is here to be read, then closed.** It carries the harness (`test/repair_information/`), the probes, the measured results and the method notes (`docs/studies/repair-information/`). Nothing here is proposed for merge.

## The question

ORFS repairs timing against two different pictures of the same wires: at CTS from `estimate_parasitics -placement`, and at global route from `estimate_parasitics -global_routing`. If the *information* is what makes the late repair good, a throwaway global route before the early one should let it decide as well — a fork/join. This measured whether that holds.

## The finding

**Better information, same ranking.** On `wirebound` in a shape where the two instruments genuinely disagree, global route is ~90× more accurate about the period (+0.8 ps against placement's +73.5 ps on a 318.0 ps routed truth), halves the per-endpoint error spread (2σ 15.0 ps against 28.8) and cuts the worst case threefold. But `repair_timing` consumes a *ranking*, and there it goes 0/10 → 2/10 on the worst endpoints, with placement marginally ahead on the worst 5 %.

The two facts are consistent, and the reason is measurable: **the worst ten endpoints span 9.4 ps and global route's own per-endpoint error is 15.0 ps at 2σ.** The ordering lives inside the instrument's error bar. That is a bound on the fork/join rather than an opinion — re-aiming needs sub-9 ps per-endpoint accuracy and a throwaway route delivers 15.

**And the repair itself barely moves the period.** Four arms, three placement seeds each, on a design that misses its constraint:

| arm | mean Δ min_period | mean Δ logic area | resolves? |
| --- | --- | --- | --- |
| `no_cts_rt` | −1.50 ps | **−30.3 µm²** | area only |
| `no_grt_rt` | +1.15 ps | **−24.0 µm²** | both |
| `no_rt` (both off) | −1.06 ps | **−65.5 µm²** | area only |

Baseline seed spread: **2σ = 19.94 ps**. So `repair_timing` costs **2.7 % of logic area** and returns a period change that does not resolve, while the placement seed is worth 20 ps — **17× the entire benefit of the repair.** Max-slew and max-cap violations are zero in every arm, so it is not buying electrical cleanliness either, and the two repairs are additive rather than redundant.

## Then: can the repair's work be undone, or reclaimed?

Three follow-ups, all negative, each closing a route the earlier
findings left open.

**Ripping the buffers out before global route destroys the design.**
271 ps becomes 4456, 5399 and 5892 ps across seeds. The hook removes
~1140 signal buffers (correctly skipping the ~163 on clock nets), but
only about 100 are the CTS repair's work — the rest are synthesis's and
`repair_design`'s max-cap and max-slew buffering. **"The early repair's
buffers" is not what is present at global route**, and the global-route
repair cannot rebuild what is taken away.

**Letting the repair choose which to remove does nothing.** With
`unbuffer` in `SETUP_MOVE_SEQUENCE` at grt, neither axis resolves
(−1.03 ps, +5.2 µm²). `unbuffer` is a *setup* move: it removes a buffer
when that improves slack. These buffers do not hurt setup, they merely
fail to help, so a repair with accurate parasitics has no reason to
touch them.

**`RECOVER_POWER`, the knob designed to reclaim exactly this, cannot.**
Twice over. It is unreachable by default — `recover_power_helper` is
called only inside `if { !$::env(OPT_POST_GRT_WNS) }` and that defaults
to 1, so setting it on the stock flow measures the gate. Opened, with
its own control for the gate, it runs (the downsizing message appears
once per arm, attributed to each arm's own `5_1_grt` action) and
produces results **byte-identical to base** on period, area and
instance count. The reason is the library set: `repair_timing
-recover_power` downsizes *and switches to higher Vt*, and this flow
reads only asap7's **RVT** liberties though the platform ships LVT, RVT
and SLVT. There is no higher-Vt cell to switch to.

So the 30.3 µm² is not reclaimable by the mechanism built to reclaim
it, **because the platform configuration denies it the cells** — which
makes the next experiment a platform-configuration change rather than a
repair-policy one.

*(In passing, a fifth empty arm: `OPT_POST_GRT_WNS=0` is byte-identical
to base, so the post-GRT WNS repair does nothing on this design either.)*

## What this rules out

- **The fork/join**, for re-aiming the early repair: the aim cannot change, because the instrument cannot resolve the band.
- **Recalibrating `set_wire_rc`**: a demand-weighted scalar moves R by 4–7 % and recovers **15 %** of the magnitude error. The error is not in the average — the router gives *critical* nets better-than-average metal, and no scalar can express a per-net quantity correlated with criticality.
- **"Congestion is not the driver"**: it is. Net length is not. The spread shapes have the longest nets, the most wire share, and the *smallest* place-to-grt gap.
- **Rip-up-and-redo**: wholesale removal is catastrophic and selective removal is a no-op. The cheapest way not to have the area is not to insert it.
- **Reclaiming the area afterwards**: blocked by a default gate, and then by the platform's RVT-only library set.

## Things the study got wrong and corrected

Kept deliberately, because each was a well-formed wrong number:

1. `-path_group reg2reg` is not available to a probe — the group is created by the *platform* SDC, a stage's written-out SDC does not carry it, and STA-0527 silently falls back to every path in the design.
2. `-to [all_registers]` then admits **in2reg** paths judged against an 80 ps `set_max_delay` with clock latency ignored. This made `wirebound` look like the *least* wire-dominated design measured (0.6 %); it is the most (26.5 %).
3. The trial route omitted `-resistance_aware`, which ORFS passes and asap7 enables. Without it the shipped `wirebound` appears to put **zero** demand on M8/M9; the flow's own route puts 6.9 % there. Two claims were withdrawn.
4. Summing every instance gave *identical* cell area for arms differing by 139 cells — filler is inserted to occupy whatever the logic leaves.
5. `sta::max_fanout_violation_count` segfaults when no fanout limit is set.
6. Summing every instance gave *identical* cell area for arms differing by 139 cells — filler is inserted to occupy whatever the logic leaves, so a total including it is constant by construction.
7. `RECOVER_POWER` looked inert until preflight found the `OPT_POST_GRT_WNS` gate in front of it. An arm that set it alone would have reported a clean, fast, meaningless no-effect.

## Limits

One design, one shape, one clock, three seeds. Setup only — hold timing is not examined. `u12_tight` was not run through the QoR ensemble. The period axis is near closure, where repair stops at zero by construction. Whether loading asap7's full LVT/RVT/SLVT range lets `RECOVER_POWER` do its job is not tested here, and is the one experiment this study points at with an obvious upside.

## Reproducing

```sh
bazelisk run //test/repair_information:update_wirebound     # the instrument ladder
bazelisk run --jobs=2 //test/repair_information:update_wbq  # the QoR ensemble
bazelisk run //test/repair_information:report -- --results $PWD/test/repair_information/results
```

`--jobs` is not optional: twelve concurrent flows exhausted a 30 GB machine.


oharboe:
Closing as intended, reference only. Self-contained: the harness, probes, measured results and method notes are all on this branch, and the write-up is in the PR body above.

Nothing here is proposed for merge. Two candidates were split out separately and are open on their own: #1012 (wirebound metal-usage documentation, off `main`) and #1019 (silicon reference appendix, off `study/coremark-joule`).

The branch is not deleted: the reproduction commands and the results it carries are the only record of the campaign.

oharboe:
## Follow-up: ripping up the repair before global route

Asked after the close, measured on the same vehicle and pushed to this branch. Two arms, three seeds each — both negative, and the premise fails before the answer does.

| arm | Δ period | resolves at | Δ logic area | resolves at |
| --- | --- | --- | --- | --- |
| `unbuf_selective` | −1.03 ps | 3.13 | +5.2 µm² | 9.4 |
| `unbuf_wholesale` | **+4974 ps** | 835 | −158.6 µm² | 12.2 |

**Wholesale removal destroys the design** — 271 ps becomes 4456, 5399 and 5892 ps across the three seeds. The hook removes ~1140 signal buffers (correctly skipping the ~163 on clock nets), but only about 100 of those are the CTS repair's work. The rest are synthesis's and `repair_design`'s max-cap and max-slew buffering. **"The early repair's buffers" is not what is present at global route**, and the global-route repair cannot rebuild what is taken away.

**Selective removal does nothing** — neither axis resolves. `unbuffer` in `SETUP_MOVE_SEQUENCE` is a setup-timing move: it removes a buffer when that improves slack. The CTS repair's buffers do not hurt setup, they merely fail to help, so a repair with accurate parasitics has no reason to touch them — even though removing them wholesale costs nothing (`no_cts_rt`: −30.3 µm², period unchanged).

So the cheapest way not to have the area is not to insert it, and rip-up-and-redo cannot beat that. The knob for recovering it from an existing flow is `RECOVER_POWER`, which runs `repair_timing -recover_power` on non-critical paths for exactly this purpose and is not measured here.

The catastrophic hook is kept on the branch: it is what establishes that the buffer population at global route is overwhelmingly not the timing repair's.

oharboe:
## Follow-up: RECOVER_POWER, the knob designed to reclaim the area

The study's last open question. The CTS repair spends 30.3 µm² for a period change that does not resolve; `RECOVER_POWER` is what ORFS ships to reclaim exactly that. It does not work here, for two reasons.

**First, it is unreachable.** `global_route.tcl` calls `recover_power_helper` only inside `if { !$::env(OPT_POST_GRT_WNS) }`, and `OPT_POST_GRT_WNS` defaults to **1**. Setting `RECOVER_POWER` on the stock flow measures the gate, not the knob. Caught by preflight before a single flow ran.

**Second, opened, it runs and changes nothing.** Opening the gate needs its own control, since it also removes the post-GRT WNS optimisation:

| arm | Δ min_period | Δ logic area | Δ instances |
| --- | --- | --- | --- |
| `no_optwns` | identical | identical | identical |
| `recover_power` | identical | identical | identical |

Byte-identical to base on every axis, across all three seeds. The knob *did* reach the tool — `Downsizing/switching to higher Vt for non critical gates` appears once per `recover_power` arm, attributed to each arm's own `5_1_grt` action — so this is not another unreachable-variable result.

**The reason is the library set.** `repair_timing -recover_power` downsizes and switches to higher Vt. This flow reads only asap7's **RVT** liberties, though the platform ships LVT, RVT and SLVT (24/25/24 files). There is no higher-Vt cell to switch to: the move the knob exists for is unavailable by construction, and downsizing alone found nothing worth doing.

So the 30.3 µm² is not reclaimable by the mechanism designed to reclaim it, on this platform configuration, **because the configuration denies it the cells**. Whether loading the full Vt range changes that is a different experiment and is not run here — but it is now the concrete next thing to try, and it is a platform-configuration change rather than a repair-policy one.

**A third negative in passing:** `OPT_POST_GRT_WNS=0` is byte-identical to base, so the post-GRT WNS repair does nothing on this design either — consistent with every other repair arm here.

PR #1029, opened 2026-09-17
tools/structured_gen: placed multi-port register files as macros, built in odb

oharboe:
A `cc_binary` that links OpenROAD's own libraries and writes a multi-port register file as a placed standard-cell macro: word rows by bit columns, every cell placed FIRM by construction, nothing left to a placer. It exists because of what a many-port register file does to a legaliser when synthesised to flops, and because a core in XiangShan's class has these as compiled bitcells rather than logic.

## Why

On XiangShan (#988), the four floating-point register-file parts (256 x 16, 14 read and 7 write ports) stopped the fp region block's detailed placement twice, at 17 k misaligned cells after 4 h 10 min and again after 1 h 17 min at lower utilisation with padding and a wider legaliser window; the integer register file (224 x 64, 12R 9W) then did the same to the 7.5 M-instance parent, 5.5 M violations to 17 k in 400 iterations and three hours, a thousand fewer per 50 after that. The residue was the register files' flops and read-mux buffers every time. The workaround in #988 makes each register file a block of its own at 20 % utilisation; that legalises in 17 iterations and costs 5.1x die over cell area. This is the fix.

OpenROAD's `ram` module is the precedent: a flop RAM generator, in the binary as `generate_ram`, allowing many read ports and exactly one write port. asap7's `fakeregfile_*` LEFs are 1R1W. Neither covers a 14R7W file.

## What it does

- Reads a small `key value` spec: words, bits, the RTL's own port names per read and write port, the cells by name. A spec is complete or refused with the offending line; a cell or pin the LEF lacks is a refusal, not a fallback.
- Builds the netlist and placement in odb: a tile per (word, bit) with the flop, an inverter for Q, the write mux as AO22 pairs and an OR2 tree, one AND2 per read port; write and read decode in a header column per word; read bitlines as balanced OR trees whose nodes run down the column in the tiles they cover; `banks` word columns side by side with a footer OR band; a service column every `tap_columns` with the tap cell and `service_sites` free sites per row, which is where CTS puts its buffers.
- Writes the ODB, a structural Verilog netlist with the RTL's bus port names, and a DEF.

## Hooked into the memories step, by name

`STRUCTURED_MEMORIES` (an ORFS variable, carried in `patches/0071`) lists spec files, one per module. Nothing is detected: XiangShan's register files are `Reg(Vec)` with `Mux1H` writes, which no yosys memory inference sees, and a `Mem` rewrite would change which writer wins; the module is named, its RTL stays as it is, and the RTL remains the simulation model. `gen_memories.py` runs the generator per spec with `--check-ports` against the Verilog file declaring the module, writes `<module>.lef`, `.lib` and `_pre_layout.lib` beside the AUTO_MEMORIES views, appends the module to `blackboxes.txt` after FakeRAM has written it, and records what it did in `structured.json`. bazel-orfs stages the generator into the canonicalize sandbox and names it through `STRUCTURED_GEN` when a design sets the variable; setting it without `AUTO_MEMORIES=1` is refused at analysis time. `tools/structured_gen/test/rf_parent` is a parent whose `rf32x32` is replaced this way through synthesis and floorplan.

The lib is a model in FakeRAM's manner, so it lands in seconds inside synthesis: loads from fanout, a combinational read arc per data bit related to every address bit, setup and hold on the write pins, constant leakage per cell; the routed block's own abstract replaces it for a measured run. `--check-ports` found the `reset` port XiangShan's register files carry and never use; it is a spec key now. `read_banked` covers Chisel's `RegfileBank`, one address and data bus per bank, which XiangShan's `IntRegFile` is: 224 x 64 in four banks of 56, 12 banked read ports, 9 write ports, 544,968 instances in 812 x 122 um, 200 s.

## How it enters the flow (as a block of its own)

No new rule. `SYNTH_NETLIST_FILES` makes the netlist stand in for synthesis, `FLOORPLAN_DEF` makes the DEF the floorplan (`read_def -floorplan_initialize`), the tapcell hook is a no-op because the tap columns are placed already, and ports are not buffered because the pins are on the die edge. Placement finds everything FIRM and legal; CTS, global route and the abstract run as on any block. `tools/structured_gen/test` carries a 32 x 32 4R2W file through that to a grt abstract in about 90 seconds, as the tool's CI.

## Numbers

| | instances | outline | time |
|---|---|---|---|
| FpRegFilePart0 256x16 14R7W, one bank | 196,808 | 54 x 554 um | 5 s |
| same, four banks | 196,808 | 232 x 140 um | 5 s |
| same as a flat block at 20 % in #988 | 101,272 | 246 x 246 um | ~15 min, legal |
| same in the fp region block at 28-40 % | | | failed twice, hours |

Twice the cells of the synthesised version (the AND-OR bitline form rather than a mux tree; regular, short, predictable) at half the die of the low-utilisation block, and nothing for a legaliser to do.

## Carried against OpenROAD

`@openroad//:openroad_lib` is package-private upstream. `patches/0002-openroad-openroad-lib-public.patch` makes that one target public, nothing else, and is carried here per the moratorium; its header says how it retires.

## Known limits, in the README

Two write ports to one word in a cycle OR their data (Chisel gives last-port priority; XiangShan never issues it; the simulation model is the RTL). No reset. Pins on one layer at the die edge in port order. One flop kind, no latch, no column mux. Stitching into #988 is the spec files plus the variable in its `config.mk`; the "before" there is recorded as how far the flat register files got before the run was stopped.


oharboe:
@maliberty cc_binary() that links with OpenROAD to generate register files to avoid watching paint dry.

PR #1038, opened 2026-09-17
coremark-joule: correctness and consistency pass, restructure, and a regenerator for Appendix A's data

ovebryne:
A review pass over `test/coremark_joule/README.md` and the data behind Appendix A: correctness and internal consistency first, then structure, then the one input that nothing re-derived.

Seven commits, each a separate claim. Nothing in the flow changes; the measured results are untouched.

### Corrections

**Figures the prose quotes now match the tables they come from.** §4.2's secondary observation printed a vectorless/measured pair whose own parenthesised ratio did not divide it — 67.65 against 44.43 is 1.52x, not the 1.23x beside it. Those figures were three drafts old while the ratios rendered from `results.json`; they now read `vectorless_power_w` against `power_w`, so dividing the printed numbers reproduces the printed ratio for all four cores. §5.1's per-core macro powers were out by about a factor of two; §5.3 and §8.6 attributed 72 % and 66 % macro to SERV and picorv32, which are picorv32's and ibex's shares.

**Nine places where two parts of the paper disagreed.** §1 said the large-core literature answers "at a stated boundary", which §4.4 and §4.5 both record that [5] does not. §2.3 named SERV as its leakage-dominated example, where SERV has the highest clock in the study and Table 1 puts its leakage at 0.00 mW. §3.7 rated SERV's attribution "excellent" while §5.7 reports it produces nothing. §5.1 called 41 million cycles "41 million accesses" four paragraphs before saying SERV has the lowest access rate.

**Three claims the measurements do not carry.** The activity sweep bounds the estimator, not the SAIF's share: `seedActivities()` iterates `levelize_->roots()` and skips leaf clock pins, and `findActivity()` sends a clock-network pin down a branch returning `2.0/clk->period()` that never reads `user_activity_map_`. So the sweep rules out the estimator — the checkable result — while a clock pin's activity comes from the SDC either way, which is why §5.3b's time-base error moved one core's total by 2.9 %. §5.4 applied the corner's V² ratio to switching power alone, which is 15 % of ibex against 81 % internal. And §4.9's ibex cross-check, the one that comes back unexplained, now records that [13] used a latch-based register file where this study's ibex hardens as flops.

**One name per thing, and references a reader can follow.** The boundary had four spellings and its complement two, mixed inside single table headers. `unmatched` meant three different things. [11] cited a path that does not exist — the document is `seh1_SweRV_CoreMark_Benchmarking.pdf` in the SweRV Support Package, and the entry now says the figures cannot be retrieved from it. [22] had no author, publication or URL while being §A.4's only corroboration; it is Beebom, and its 47.6 W is a peak during Geekbench 6, so §A.4 now claims the clock and not the power. Tables are numbered in the order a reader meets them.

### Structure

Threats to validity was the largest section in the paper, half again the size of Results, and carried three things that are not threats.

The boundary correction — the abstract calls it the study's largest result — is now **§4.2, in Results**. §5.1 keeps what is genuinely open. **§6, "What the flow got wrong"**, carries the five findings that are defects in the flow rather than in this measurement: the SAIF time base, attribution on parameterized modules, the 80 ps IO default, VeeR's clock gates, the two carried workarounds. Each applies to designs with nothing to do with CoreMark, and §6.3 says so of itself.

**§9 is a conclusion** — the central finding was argued in §4.6 and never restated. **§10.3 is what a core has to supply to join the study**, which was one clause in §1 and nothing else. Appendix A is one section with two methods instead of two arguing about which is which; Further work is one section; the limitations list is stated once, after the abstract; there is a contents list.

### Appendix A's data

The table showed 15 of the 22 rows in `commodity_coremark.csv`, and the prose compared against one it left out — "Xeon 8490H against Core i9-13900K is 1.26x" named a part no reader could find. The omissions that carry an argument are now in, and `scripts/fetch_commodity.py` regenerates the CSV from the four OpenBenchmarking exports, with `--check` failing on drift.

The measured numbers come from the export every run. What it does not have is carried in `SELECTION`, because there is no rule to derive — 61 system entries become 27 rows and the omissions are a choice, so the choice is in one reviewable place. Running it surfaced that `reported_ghz` is inconsistently sourced and that `power_min_w`/`power_max_w` are absent from the CSV export entirely; the script reports both rather than silently resolving them.

**The exports are Phoronix's data and are deliberately not committed.** `--check` is a local step, and the unit tests cover the parser, the renderer and the manifest. §11 now says where Appendix A's rows come from; it covered CoreMark and nothing else.

Adding the parts also corrected §A.6: it said the package figures "span 4,827 to 12,028" and that the three wall-plug numbers "land inside that band", while the Xeon's 4,700 was below the floor.

### Two things left open

- **Not everything in the exports is in the table.** The AMD desktop block is now the complete Zen 4 and Zen 5 X-series with each line's 65 W part, but the Zen 3 parts, the X3D and APU variants and the Core i5s are still out with no stated rule. `SELECTION` is where that would be written down.
- **The Zen 5 dual-die parts do not behave** and §A.1 reports it rather than dropping the rows: the 9900X and 9950X regress 14.1 % and 21.6 % against the 7900X and 7950X, while the single-die 9600X and 9700X improve by 10.3 % and 6.1 %. What splits them is not established here.

### Checks

Every commit in the branch was extracted and tested on its own: `section_links_test` and `readme_numbers_test` pass at all seven. The final document has no duplicate section numbers, no dangling `§` references, no broken anchors, and tables 1–15 in document order. The whole `scripts/` suite is 252 passed; the 3 `iteration_delta_test` failures need Bazel-built ELFs and fail identically on unmodified `main`.


ovebryne:
@oharboe — for your review. I could not set the reviewer field: this comes from a fork and I only have `pull` on this repo, so the review-request API returns 404.

Two things that may be worth looking at before the rest:

- **`test/coremark_joule/scripts/fetch_commodity.py`** regenerates `results/commodity_coremark.csv` from the four OpenBenchmarking exports. I deliberately did **not** commit the exports — they are Phoronix's data, ~343 kB of complete result sets, and committing them into a BSD-3 repo would offer them under a licence we cannot grant, quite apart from what DCO sign-off would be certifying. The consequence is that `--check` is a local step rather than a CI one. If you would rather have the exports in-tree under their own NOTICE, that is a permission question for Phoronix and I can prepare it.

- **`SELECTION` in that script is the part that needs your judgement.** There is no derivable rule behind the commodity table — 61 system entries across the four exports become 27 rows — so I wrote the choice down rather than inventing a rule for it. The AMD desktop block is now the complete Zen 4 and Zen 5 X-series with each line's 65 W part, but the Zen 3 parts, the X3D and APU variants and the Core i5s are still out with nothing stating why.

ovebryne:
Rebased onto current `main`, which had moved 13 commits and added `5.2c A second core, and where that stops` and `8.8 Glitch power per unit, and per core` to this README.

Both are carried over word for word. Only their numbers changed, to fit the restructure: **5.2c is now 5.4** (after 5.3, the multiplier measurement that was 5.2b), which shifts the old 5.4–5.10 down one to 5.5–5.11; **8.8 is now 8.9**, since 8.1 is the roadmap of cores that Further work absorbed. Their table is now **Table 9**, and the rest renumber to keep the sequence in document order.

Their internal references were remapped with everything else: `§5.2b`→`§5.3`, `§5.7`→`§6.2` (attribution, now under "What the flow got wrong"), `§7`→`§8.1`, `§4.9`→`§A.1`.

I diffed both sections against `main` after the merge: the only differing lines are the heading and the table caption. Every § reference resolves, no duplicate section numbers, no broken anchors, tables run 1–16 in document order, and `scripts/` is 263 passed (the 3 `iteration_delta_test` failures need Bazel-built ELFs and fail the same way on unmodified `main`).

PR #1040, opened 2026-09-18
patches: SKIP_EXTRACT_FA skips yosys's full-adder extraction, a turnaround knob

oharboe:
`SKIP_EXTRACT_FA=1` leaves out yosys's full-adder extraction and lets adders map through the generic techmap. Extraction is a large share of synthesis time on arithmetic-heavy modules: an FMA or a vector converter can spend most of its partition in `extract_fa`. A turnaround knob for architectural exploration, off by default. Carried as ORFS patch 0074; retires at a `//:bump` onto an ORFS that carries the knob.

Stacked on the auto-memories PR. Split out of #988.


oharboe:
@maliberty The cat dragged this in: extracting full adders in yosys is slooow.

PR #1048, opened 2026-09-18
odb-debug: a persistent OpenROAD session on a stage ODB, for questions

oharboe:
A stage of the flow ends in an ODB, and everything about where the flow left the design is in it. Reading it back costs the ODB load and, for timing, the liberty set, every time a script opens it to print one number; on a million-instance design that is minutes per question.

`odb_debug(name, src = ":cpu_place")` is a runnable target that opens the stage's ODB in the flow's own OpenROAD, with the flow's SDC, liberty, RC and derates, and keeps it open behind a localhost socket until it has been idle for a while. `tools/odb_debug/daemon.tcl` is the session (JSON-answering `od_*` procs for status, macros, cells, nets, legality, geometry dumps, worst paths; base64 Tcl in, JSON out; `daemon.json` with port, ODB, workspace commit); `odbdebug.py` the client; `mcp_server.py` an MCP server exposing the procs as tools, registered in `.mcp.json`; `geometry.py` the placement forensics a failed legaliser asks for. `GUI_TIMING=0` skips liberty and a 1.5 M-instance floorplan loads in 37 s. The same script runs on any checkpoint ODB from a deployed `_deps` tree.

Standard library, Python 3.6 syntax, declared in `host_tools.py`. Tests: client and server against a fake daemon, the forensics on a hand-made dump, and a smoke test that launches the real daemon on the placed multiplier. `/odb-debug` is the skill; `docs/debugging.md` points at it.

Written for #988, where it found that the parent's legaliser was failing on 22 000 row fragments narrower than 20 µm between abutting macros.


oharboe:
@maliberty MCP server for debugging odb

PR #1052, opened 2026-09-18
orfs: one global placement with -place_ios when neither driven mode is on

oharboe:
The place stage runs global placement twice: `3_1_place_gp_skip_io` ignores the IO pins, `3_2_place_iop` places the pins against that result, `3_3_place_gp` places again with the pins known. With `GPL_TIMING_DRIVEN` and `GPL_ROUTABILITY_DRIVEN` both off, the second solve repeats the first with pin awareness as its only difference. OpenROAD's `global_placement -place_ios` co-optimises the movable pins with the cells in one solve and leaves them for `place_pins` to legalise; it refuses the two driven modes, so it fits exactly the case where the second solve has nothing to add.

Carried ORFS patch 0077, automatic when both modes are off: `3_1` runs the full global placement with `-place_ios` (port buffering and dont-use included) and leaves a marker file; `3_2` legalises the pins as before; `3_3` finds the marker and skips its `global_placement`, keeping its parasitics estimate, flop clustering and metrics. The three checkpoints and their names are unchanged, so nothing downstream moves. `FLOORPLAN_DEF`, `FOOTPRINT` and pre-placed pins keep their existing paths; with either driven mode on, nothing changes.

Measured on the anchored wirebound arms of the global-route scaling study (turnaround settings, both driven modes off):

| arm | place stage before | after | HPWL after 3_3 | I/O-net HPWL at 3_2 |
|---|---|---|---|---|
| 500 µm die | 3:22 | 2:09 | 1.13 M → 1.07 M µm | 1.06 M → 0.89 M µm |
| 2 mm die | 9:02 | 5:17 | 4.24 M → 4.05 M µm | 4.18 M → 3.85 M µm |

Pins placed in the solve beat pins placed after a pins-blind solve. The XiangShan parent, whose second solve was 17 minutes, runs with both modes off and picks this up at its next build. Not upstreamed; retires at a `//:bump` onto an ORFS whose place stage places pins with global placement.


oharboe:
CI was failing for two reasons, neither of them this branch's subject; both fixes are now on the branch.

1. `test/coremark_joule/sim:xiangshan_rv64gc_crc_test` was the one CRC test without `tags = ["manual"]`, so `bazelisk test ...` in CI built the Verilated XiangShan SoC and tried to simulate 50 M cycles. Every run that could not serve it from cache died mid-suite with "the hosted runner lost communication with the server" — main's runs included. Tagged manual, like every sibling simulator, run and CRC test; the CI suite goes 217 → 216.

2. `.bazelignore` still carried a bare `.claude` next to `.claude/worktrees`, which left `//:guard_tool_test` declaring sources bazel would not stage: it built and then died on exec with "file not found" for its own main file.

main carries both still; it goes green when one of these merges.

oharboe:
## Closing: `-place_ios` is not ready to carry a flow on

CI on the first push died in `3_1_place_gp_skip_io` with

```
[ERROR GPL-0305] RePlAce diverged during gradient descent calculation,
resulting in an invalid step length (Inf or NaN).
```

on `sram/sdq_17x64`. Two separate numerical defects in gpl's concurrent
IO placement, both reachable only through `-place_ios`, both reproducible
from a floorplan `.odb` in three lines and a couple of seconds:

**1. The step-length norms lose the cells to rounding.** `-place_ios`
adds the movable pins to the cell list as virtual GCells; gpl keeps them
out of the step-length norms by summing *every* GCell's squared
displacement and subtracting the IO terms afterwards. A pin crossing the
die and a cell moving a fraction of a site are orders of magnitude apart,
so in `float` the cell terms vanish into the rounding of the IO sum and
the subtraction can leave a small negative residue. `sqrt` of it is a
NaN step length, reported as divergence:

```
[DEBUG GPL-getStepLength] CoordinateDis 7.18432, GradientDist 0.00185616
[DEBUG GPL-getStepLength] CoordinateDis -nan,    GradientDist 0.138467
```

Carried as `patches/0004-openroad-gpl-place-ios-nan-step.patch`: skip the
IO terms rather than subtract them.

**2. A stopped solve is reported as a diverged one.** The step length is
the coordinate distance over the gradient distance. A gradient distance
of exactly zero — the gradients stopped changing, nothing left to move —
divides out to an infinity, and every non-finite step length is read as
divergence:

```
[DEBUG GPL-getStepLength] CoordinateDis 25.782, GradientDist 0, StepLength: 878.293
[ERROR GPL-0305] ...
```

Carried as `patches/0005-openroad-gpl-zero-gradient-stop.patch`: finish
the region instead, with a warning naming the overflow it stopped at.

With both patches the two mocked SRAMs place and the full suite is green
(216/216 locally). But the second failure is a symptom, not a bug on its
own: on `sdq_17x64` mocked the solve runs **2553 iterations without
converging**, overflow stuck around 0.43 and the density penalty climbing
to 1e33, before it stops moving at all. 16 movable instances against 40
movable pins is a solve with nothing to anchor it — the pins follow the
cells they pull. That is the shape of a feature that has not met small
and degenerate designs yet, and `patches/0005` papers over the end of it
rather than fixing it.

So this stays unmerged. The two gpl defects are worth reporting upstream
on their own, with the reproducers and the patches; the flow change waits
for a `-place_ios` that survives a design like this one.

The measured wins in the description stand and are the reason to come
back to it: place stage 3:22 -> 2:09 and 9:02 -> 5:17 on the anchored
wirebound arms, with the I/O-net HPWL `place_pins` reports dropping 16 %
and 8 % — pins placed in the solve beat pins placed after a pins-blind
one.

Branch kept for reference.

PR #1053, opened 2026-09-18
bump: ORFS 95ebc50a2583, OpenROAD HEAD d35e70836dd4

oharboe:
`bazelisk run //:bump -- --head=openroad`, plus the patch work the bump demanded.

## Retired — upstream carries them now

| patch | upstream |
| --- | --- |
| `0070-orfs-detail-route-metric-integer` | ORFS #4542 — the commit this bumps onto |
| `0004-openroad-write-pin-placement-brace-names` | OpenROAD #11437 |
| `0005-openroad-tcl-full-command-names` | OpenROAD #11439 — the commit this bumps onto |

0005 was checked by hand rather than by a reverse-apply: `synthesis.tcl` has drifted around the hunks, but every call site the patch spelled out (`syn::abc_roundtrip_cmd`, `syn::map_sequentials_cmd`, `syn::map_combinationals_cmd`, `syn::export_to_odb_cmd`) is upstream.

## Rebased — real conflicts

- **`0037-orfs-single-writer-1_synth-sdc`** — `synth_odb.tcl` grew a LEC/SEC `write_lec_verilog` block next to the sdc write. The unconditional `write_sdc` keeps its place and the new block follows it.
- **`0075-orfs-variables-stages-complete`** — `docs/user/FlowVariables.md` is generated from `variables.yaml`, so the merged yaml was fed back through `flow/scripts/generate-variables-docs.py` rather than hand-merged. The result is byte-identical to what the generator emits from the patched yaml.
- **`0001-openroad-tcl-source-flag`** — `Main.cc` gained `web_bind_arg` beside `web_port_arg`; `tcl_source` follows it in both the declaration and the `findCmdLineFlag` block.

## The rest is regeneration, not new content

Every carried patch is now written with exact hunk offsets against the new pin. Two reasons:

- bazel's `ctx.patch` is stricter than `git apply` and rejected hunks that git placed with an offset;
- the GNU `diff -ruN` separators some patches used lose the file boundary once a hunk header shifts, so a later hunk lands in the wrong file.

The whole series is now `diff --git`, no `index` lines, no timestamps.

## Rebased onto main

`main` landed the coremark-joule rewrite while this sat in CI. That rewrite
already links every `§` reference, so the README commit this branch carried is
gone: `section_links.py` finds nothing to fix against the new headings, and the
anchors the old commit wrote named the old ones. Three commits remain — the
`.bazelignore` one-line removal, the bump, and the `USE_FILL` stage fix — and
everything outside `test/coremark_joule/` is byte-identical to what CI last
analysed.

## Verification

- Applying all 29 ORFS patches in `ORFS_PATCHES` order onto `95ebc50a2583` yields a tree identical to the rebased series.
- `@orfs` and `@openroad` both fetch and patch clean.
- `//:fix_lint`, `//:public_surface`, `//:host_tools` all pass.
- Full local suite on the rebased branch: **217 of 217 tests pass**
  (`bazelisk test ... @orfs//flow/designs/asap7/uart:uart_test --build_tests_only --keep_going`),
  the uart test included — that is the one the `USE_FILL` fix was for.


oharboe:
CI's `Run tests` step failed on `//:guard_tool_test`, and it is not the bump:

```
AssertionError: Cannot exec() '.../guard_tool_test.runfiles/_main/.claude/hooks/guard_tool_test.py': file not found.
```

`.bazelignore` carries both `.claude/worktrees` and a bare `.claude`. 3b689297 narrowed the entry to the worktrees *precisely because* `.claude/hooks` holds the sources of `//:guard_tool_test`, and an ignored package leaves that target declaring srcs bazel will not stage — it builds, then dies on exec with "file not found" for its own main file. 03818962 put the bare line back underneath the narrowed one.

main has been red on this since 03818962; the last green run on main is 2026-09-17T13:02. The one-line removal is now the first commit on this branch, ahead of the bump.

oharboe:
Full local run: **216 of 216 tests pass** (`bazelisk test ... --build_tests_only --keep_going`).

The run before the fixes had three failures, two of them older than this bump:

- `//:guard_tool_test` — the `.bazelignore` regression above.
- `//test/coremark_joule/scripts:section_links_test` — seven bare `§` references in the coremark-joule README from 020852b4 and b00e11a6. Fixed with the repo's own `section_links.py --fix`.
- `@orfs//flow/designs/asap7/uart:uart_test` — failed to build:

  ```
  no such variable  (read trace on "::env(USE_FILL)")
    (file ".../flow/scripts/density_fill.tcl" line 6)
  ```

  `density_fill.tcl` opens with `erase_non_stage_variables final` and reads `$::env(USE_FILL)` at file scope, but patch 0075 scoped `USE_FILL` to `route` — so the one stage that reads it is the one stage where it is unset, and any design reaching `6_1_fill` dies there. `density_fill.tcl` says `final` on both sides of the bump, so this predates it.

  Rather than fix only the case the suite caught, I audited the rest of 0075: every script's `erase_non_stage_variables` stage against the variables it reads, with guarded reads and reads inside procs the stage never calls separated out. No second case — `RECOVER_POWER` is reached only from `recover_power_helper` (global_route only), `REPORT_CLOCK_SKEW` sits behind a short-circuiting `$include_clock_skew`, and `PLACE_DENSITY` / `CELL_PAD_IN_SITES_GLOBAL_PLACEMENT` come from `place_density_with_lb_addon` (place only).

Branch is four commits, smallest diff first: `.bazelignore` → README links → bump → `USE_FILL`.

oharboe:
Rebased onto `main` (`2c8229ca`), three commits now: `.bazelignore` → bump → `USE_FILL`.

The README commit is **dropped**, not merged. `main`'s coremark-joule rewrite
already links every `§` reference — `section_links.py` reports nothing to fix
against the new headings — and it renumbered sections, so the anchors that
commit wrote name headings that no longer exist. Resolving the conflict either
way would have put stale anchors back.

Everything outside `test/coremark_joule/` is byte-identical to the branch CI
last analysed (`git diff <old head> <new head> -- ':(exclude)test/coremark_joule'`
is empty), so nothing about the bump itself changed.

Local on the rebased branch: `//:fix_lint`, `//:public_surface`, `//:host_tools`
pass, and `bazelisk test ... @orfs//flow/designs/asap7/uart:uart_test
--build_tests_only --keep_going` is **217 of 217 passing**, uart included.

oharboe:
CI was dying mid-suite with "the hosted runner lost communication with the server", not on anything this bump touches: `test/coremark_joule/sim:xiangshan_rv64gc_crc_test` was the one CRC test without `tags = ["manual"]`, so `bazelisk test ...` built the Verilated XiangShan SoC and tried to simulate 50 M cycles on a 4-core runner. Now tagged manual, like every sibling simulator, run and CRC test; the CI suite goes 217 → 216.

The other half of the breakage, the stray `.claude` in `.bazelignore` that left `//:guard_tool_test` without its own main file, this branch already carried.

PR #1055, opened 2026-09-18
study(grt-scaling): XiangShan's parent through global route; macros chosen by interface

oharboe:
Draft, the working record of the XiangShan global-route campaign; single-concern splits follow as each part settles (the list at the end says which). Updated 2026-09-20 with the plan the branch is now executing, so this page can serve as the input to a discussion.

## TL;DR

**Have.** XiangShan's parent has never been through global route on asap7, and four takes say why. Global route on the 3.6 mm die was seven hours into maze iteration 1 of 30 at 105 GB (take 13). Pin-fitted mocks brought the die to 2.27 mm and CTS to 6 minutes (take 16), and every knob arm of the overnight matrix then timed out in the same maze iteration. A margin sweep with the pin-limited blocks dissolved (take 17) reached the zero-iteration route on every arm, and every arm hit the same wall: FastRoute's own 100×-capacity edge guard (GRT-0228) on a GCell edge a few microns from a macro's pin side, or, when the walls were avoided at the tightest margin, a die too dense to route at all (11.85 M overflow). The blocks were chosen by synthesis time, which selects the modules with the widest interface per side; the router died on those sides.

**Doing.** Choosing the macros by interface instead, and planning the floorplan their interfaces dictate, as a skill (`/macro-selection`, `tools/macro_select`, `test/macro_select`) dogfooded on XiangShan. A global route that converges is the proof; there is nothing to review before it, so the chain below runs unattended to that result and the period at global route becomes the project's KPI.

## The pivot (2026-09-21)

Takes 19 to 21 put the planned parent through its gates and stopped in the legaliser after CTS, twice. The odb-debug density map of the placement names the cause: the reorder buffer, 256 copies of `RobEntryCell(idx[8:0], in[3618:0]) -> entry[51:0]`, a 3 619-bit bus broadcast to every entry, 1.04 M cells for 13 000 flops of state, half of the densest bins; dispatch's and the memory-control tables the rest. These are compiled structures in every real core of this class and synthesised flops here. The plan, with its literature and rules, is `ideas/structured-macros.md` with `ideas/structured-macros.png`: `tools/structured_gen` grows set and clear-after ports, a Chisel patch in the fork extracts the ROB's array, one equivalence check per change, the parent rebuilt with the ROB and the register files as generated macros, then the next arrays by the same measure. First step landed the same day: XiangShan patch 0008 extracts the ROB's entry payload into `RobEntryFile`, a register-file module the flow hands to `structured_gen`, and the planned parent hardens `IntRegFile`, `RenameBufferFile` and `RobEntryFile` as generated macros; take 22 runs it on a plan sized from the parent's measured cells at density 0.5.

## Slang, always (2026-09-21)

None of XiangShan's flows had named a synthesis frontend, so all of them parsed the flat core with yosys's own, seven minutes of canonicalisation per synthesis and again after any input change (inventory entry 7). `orfs_flow` now defaults `SYNTH_HDL_FRONTEND` to slang and refuses another frontend without a `yosys_frontend_reason` next to the setting; every flow package in the repo analyses under the rule.

## Take 22 and the workbench (2026-09-21)

Take 22 ran the pivot's first step: the parent with `RobEntryFile`, `IntRegFile` and `RenameBufferFile` as generated macros, on a plan sized from the parent's measured cells at density 0.5 with plan-placed pins (every partner pin at an exact track coordinate, since the pin placer's annealer emptied short interval constraints). Blocks and parent synthesised (the parent's last partition, an issue queue with a busy table, spent two hours in abc; the hierarchical `link_design` seven minutes, inventory entry 6); the floorplan then sat in `report_metrics` and was stopped. Two changes follow from it. A generated array in the middle of the logic is now a **placed netlist** (ORFS patch 0078, `mode netlist`): blackboxed in synthesis, linked when the design loads, its cells dropped FIRM onto the parent's rows at floorplan where the plan says, so the parent keeps one placement, clock tree and route across it; macros are for terminal blocks at the edge only. And the floorplan is iterated in a deployed `_deps` tree with `./make do-<stage>` (the wrapper now refuses bare stage targets), slack read from the stage ODB with odb-debug, and the settled plan ported back so bazel reproduces the take.

## Take 23: three holes and a bug (2026-09-21)

The first take 23 synthesis on slang found three holes in the flow and one bug in the plan, in that order.

- **Chisel's verification layer.** firtool emits every assertion as a `<Module>_Verification_Assert` module bound into its parent with hierarchical references (312 in XSCore); slang refuses a hierarchical reference across a preserved module boundary, and yosys's frontend only ever dropped them silently. Fixed at elaboration: `--default-layer-specialization=disable` on the generator's firtool arguments, so every view, the simulator's included, is one Verilog without the assertions. The flat core has only that one layer and no DPI.
- **Memory extraction, twelve of fifteen minutes.** `memory -nomap` in `extract_memories.tcl` walks every module in each pass, and yosys's `memory_dff` builds its per-bit driver and consumer index for a module before asking whether the module has a memory (memory_dff.cc:225). The flat core has 1 870 modules and 407 hold a memory cell, all small firtool `ram_*` modules; the index over the other 1 463 is the cost. Measured on the flat XSCore through slang: 721 s unscoped (memory_dff 462 s) against 64 s scoped to those 407 modules, the same 407 memories. Carried as ORFS patch 0079 (inventory entry 8); the proper fix is an early return in yosys.
- **The parent still synthesises what the generators left.** Per module, one single-threaded ABC call is two thirds of each: Rob 995 s (352 `RobEntryCell` instances on a broadcast bus and the commit and walk selection over them; the cell keeps 30 bits, the file 21), the fp region 727 s and 8.4 GB (datapath and bypass muxes, four fp issue queues), Frontend past seventeen minutes when stopped (Ifu, IBuffer), Rename 217 s (RAT snapshots), SSIT 176 s, the int region 185 s. The generators have not gone deep enough in the ROB's payload and pointer-window reads, the bypass networks and the issue queues; IBuffer and Sbuffer are plain extractions still to do.
- **The bug.** `plan_floorplan.py`'s emit reused the macro row list for a block's exact pin rows and skipped the block's entry, so the plan reached the flow with `"macros": {}` and `place_macros.tcl` was Region_1's 5 099 pins under a three-field foreach. Takes 22 and 23 therefore synthesised the whole core in the parent, blocks included; the block-level tables stand, since the blocks ran as their own flows. Fixed with a test that a block with exact pins is in both files; the plan is re-emitted with its four blocks and take 23 relaunched.

## Before the next synthesis from the top (2026-09-21)

Two things to do before anyone re-runs synthesis from the top, written up so they can be implemented on another machine while take 23 runs:

- **One canonicalise session per flow** (`ideas/canonicalize-slicer.md`): today one yosys session per kept module re-reads the whole 4.9 GB checkpoint to write that module's slice, 109 sessions per synthesis here, 2.5 CPU hours and the memory wave that caps `--jobs`. The fence it guards (byte-stable per-partition inputs, commit f3c52546) survives a single session writing every slice with `write_rtlil -selected`; primitives verified, plan and test list in the file. Compute the module walk explicitly; yosys selection expansion is fickle.
- **Generator depth** (`ideas/structured-macros.md`, "Generator depth backlog"): the per-module ABC table from take 23 and what a generator should own in each. The store queue (payload file, address CAM, window read) is the clearest case; Rob's remaining glue is 6 417 pointer comparators (inventory entry 9); Bpu, Ifu, decoders and CSRs are real logic with nothing for it.
- **Correctness** (`ideas/structured-macros.md`, "Correctness: kepler-formal"): LEC per generated structure with kepler-formal wired up in bazel on the latest upstream, flops named after the RTL registers; eqy does not scale.

| parent | take 13, 3.6 mm | take 16, 2.27 mm | take 17, margin 1.5 | take 17, margin 2 (channel floor) | take 17, margin 3 |
|---|---|---|---|---|---|
| macro area | 4.15 mm² | 0.83 mm² (pin-fitted mocks) | pin-limited blocks dissolved | same | same |
| die | 3.6 mm | 2.27 mm | 1.43 mm, 40 % fit | 1.73 mm, 40 % | 1.89 mm, 50 % |
| CTS stage | 31 min, 59 GB | 6 min, 27 GB | 7 min | 42 min | 52 min |
| global-route grid | 6 337² | 3 978² | 2 505² | 3 032² | 3 318² |
| route, 0 iterations | >7 h in iteration 1 | timed out, 2.5 h, every knob arm | ok: 45 % usage, 11.85 M overflow, max edge 626 | GRT-0228, 5 µm below a pin side | GRT-0228, 7 µm above a pin side |

Take 17 without the channel floor: CTS 82 minutes, then GRT-0228 with 3 313 wires on one edge beside two IssueQueue mocks. The 34 blocks chosen by synthesis time put 6.9 pins on every micron of their pin sides; that is the number the whole campaign turns on.

## The plan

The procedure is the skill's text (`.claude/commands/macro-selection.md`); the campaign runs it once on XiangShan:

1. **Space table** (`select_table.py`): interface bits per module from the flat RTL, pins per micron of the square each module's area makes, each parent against the sum of its children. The architecture drew the narrow cuts: the whole Frontend is 3.3 k bits where its five pieces bring 17 k. A well-cut unit is under about 2 pins per micron; the out-of-order core's issue and control run 4 to 9 and never become a macro. The RTL is not changed: the design has taped out, its cuts are where they are, the job is to find them.
2. **Time table** (`probe_boundaries.tcl`, `probe_paths.tcl`, `timing_table.py`): the whole core synthesised flat with every candidate kept and `OPENROAD_HIERARCHICAL=1` at 800 ps (1000 ps at global route means 800 at synthesis), opened through `odb_debug`, `repair_design` first (synthesis-stage slack is a fanout artefact: the worst path was −7.2 ns from two unbuffered nets of fanout 1410 and 1056), then per module: are the boundary pins registered right inside, the worst path across the boundary with depth and slack, the empty pipeline stages inside, which are the retiming idiom. A second synthesis keeps every module of 1500 or more generated lines (245), to find registered boundaries wherever the RTL has them. Both targets are plain `orfs_flow()` rules now, so the data stays reproducible.
3. **Calibration** (`test/macro_select/calibrate.sh`): a synthetic parent with one pin-wall block, one flop per pin on the parent side, routed in seconds per point, gives the constants the planner cannot derive. Result (route-0, two pin layers, 13.9 pins per micron of the pin side): with the logic *facing* the pin side, every arm from 1 000 to 10 000 pins across 10 to 100 µm channels routes clean, total overflow at most 31 edges, largest edge 15 over. With the logic *beside* the pin side, every wire turning along the block, 1 000 pins overflow 2 k, 3 000 pins 1e5 with edges over 100, 6 000 pins 1e6 with edges in the hundreds, and a 100 µm channel is no better than 10 µm. The pin pitch was never the constraint; the fraction of a block's pins whose logic is in front of them is. The walls of takes 16 and 17 were lateral traffic.
4. **Planner** (`plan_floorplan.py`): from both tables, per macro the pin side (never shorter than the pins need at the calibrated pitch; the block square when they fit; two adjacent sides only past the aspect cap), the channel in front of it (wires that run along it plus a floor for the cells the parent puts there), the parent's logic region at a chosen density, the side assignment that minimises the die over all assignments, the die, and each crossing's slack after the wire it now has to cross. A crossing that goes negative names a wrong cut: that macro moves one level down the table, the floorplan is not bent around it. Every block is R0 with its pin side chosen in its own frame, so no flip and no track trouble.
5. **Build with the gates on**: blocks as real flows with the planned outline and pins on the planned side at real spacing, each synthesised at the target period so a block that cannot close alone is known before the parent is built; `SYNTH_RETIME_MODULES` per module where the idiom is present (registers back to back behind a multiplier and the like), measured per module, kept where it pays; the parent from the plan; then the gates, cheapest first, each a hard stop: the annealer's channel check at floorplan, the free-site picture and the RUDY map at place, the legaliser's window check, the zero-iteration global route (total overflow under about 100 k, no edge in the hundreds) before any maze budget is spent. A wall on a block's side names the block; dense everywhere names the utilisation.
6. **Proof and KPI**: a global route that converges within its iteration budget in bounded time; from then on the period at global route, and a timing inventory (`ideas/xiangshan-timing.md`) as the backlog that moves it, each entry a well-studied problem fixed in the tool or articulated in the RTL and re-measured on the same floorplan. Expect outliers in period; a 30 % gap to a commercial flow is the working assumption, the inventory is how it shrinks.

Rules: interfaces before areas; tables before flows; calibration on the synthetic before a take of the real design; never cut through the tangled core to make blocks build faster; a gate that fails names the stage where the fix belongs.

## Hard stops found on the way

Every one of these is a case where the flow ran for hours to report what a check could have said in a second.

- **Skills** (#1056, merged): global route is the router, not a diagnostic; the fix for a route that will not converge is usually in the stage before (odb-debug the placement: free sites, RUDY) or further back (the floorplan, the choice of macros, dissolving or adding one). The odb-debug skill now leads with a mandatory ladder: RTL, macro set, floorplan, placement knobs, stage knobs, tool code, in that order.
- **Legaliser** (patch 0004, carried): `detailed_placement` with a diamond window of `-max_displacement {450 450}` (microns) is a 8 333 × 1 666 site search per cell; a stage that took 23 s at the default window took 7 minutes, and 82 minutes on a denser parent, before failing. The patch checks Hall's condition (demand against free sites in each window) once up front and refuses a placement no search can legalise. The `-max_displacement` value itself is the user saying this hill is worth dying on, so no new flag. Upstream pull request with a gtest later, on order.
- **Global route** (OpenROAD issue #11452, filed on order): a pin-escape lower bound at the start of global route, an error by default with an option to route anyway, so that a macro whose pins cannot leave its side is named before the maze budget is spent. The mock-area sizing and the annealer's channel check on this branch are the same test on the flow side.
- **Mock pins off the parent's tracks** (fixed here): one DRT-0073 per macro instance in `pin_access`; a 1-second reproducer settled it as the origin lattice (asap7: 0.144 × 2.16 µm with pins on M2 to M5) and illegal flips (MX legal iff the height is 1.08 mod 2.16; MY impossible with M3 and M5 on one edge). Origins snap, the chooser takes only legal flips, a pin-fitted mock's side rounds up to the first legal size.

## What this branch adds

- **`test/grt_scaling`**: an IO-anchored wirebound on a swept die, `grt_bench.tcl`/`grt_bench.py` (one measurement per arm under a systemd scope with memory cap and timeout, resumable), `report.py`. Global route measured in tens of seconds; the matrix driver ran takes 16 and 17 unattended.
- **`mock_area = "pins"`**, `block_mock_area`, `MOCK_AREA_PIN_EDGES/MARGIN` scoped to the mocked variant: a block's mocked outline fitted to its pins on two adjacent edges, MX-legal sizing, never larger than the real block.
- **`macro_anneal`**: lattice snapping from the track patterns, legal flips only, escape need per side from the pin count and track density, per-side halos, channel shortfall check with an automatic widening and a floor (`ANNEAL_CHANNEL_CHECK/AUTO/MIN_UM`), two-pass shelf packer.
- **`tools/macro_select`**: `select_table.py`, `plan_floorplan.py`, `timing_table.py`, the two probe scripts, tests. **`test/macro_select`**: the pin-wall calibration harness (generated block, one flop per pin, block placed on the lattice, straight and lateral arms, `calibrate.sh`).
- **Skill**: `/macro-selection` (command and skill file), the odb-debug ladder, CLAUDE.md pointer.
- **XiangShan**: plain `orfs_flow()` targets in `xiangshan.bzl` (the config.mk DSL retires here); `XSCore_hier` (138 kept modules) and `XSCore_hier_all` (245) at 800 ps with `OPENROAD_HIERARCHICAL=1`, each with an `odb_debug` target; 40 µm channel floor; pin-limited blocks dissolved.
- **Patches**: 0077 (ORFS; one global placement with `-place_ios`, and a nearly-empty-design fallback to `place_pins` when it diverges), 0004 (OpenROAD dpl window supply check). Both carried, neither upstreamed; both are candidates.
- **Rules**: serial synthesis and the keep pass no longer remake the staged AUTO_MEMORIES artefacts (`--old-file`).

## Breakout candidates

`test/grt_scaling` (study); `mock_area = "pins"` + `block_mock_area` (one feature); `macro_anneal` lattice, flips and channels (one); `tools/macro_select` + `test/macro_select` + the skill (one, after a route converges, with XiangShan as the worked example); patch 0004 (one, with its gtest, on order); the XiangShan commits stay with the study.

## Running now

`XSCore_plan`: `xiangshan_flow(plan = PLAN, variant = "plan")` builds only the plan's blocks as real flows at the planned outline with every pin on the planned side and no mock, and the parent at the planned die with the plan's `place_macros.tcl`, without the 450 µm legaliser window. The first plan (Frontend, MemBlock, VecRegionModule, Region_1; die 1.8 × 3.6 mm, 67 % utilisation) is a smoke test of the wiring at guessed constants until the tables replace it; the Frontend block's synthesis is running. Still to land: `repair_design` on the 138-module synthesis for the time table, the 245-module synthesis, the planner on both tables, take 19 through its gates to the 5-iteration route. The result lands here when it exists.


oharboe:
@jhkim-pii FYI, I'm trying to get XiangShan through global route, but I'm not there yet. It is a BIG design :sweat:

oharboe:
## Status, take 23: what XiangShan XSCore tells us about large designs in the flow

Every number below is in `docs/images/xiangshan_take23_status.json` on this branch, which also renders the figures, or in `ideas/xiangshan-timing.md`, entries 1 to 19.

### Four conclusions first

the record

1. **Large designs checked into the flow repositories have failed every time, on the record.** ORFS retired MegaBoom on 2023-12-01 with the commit message "24+ hours designs need something more than just ORFS"; the MegaBoom repository that followed is archived; bazel-orfs's own `gallery/` of twelve designs, the largest stuck at global route, was removed in August. XiangShan's XSCore is 4.3 M cells after synthesis, larger than any of them.
2. **XSCore is not something a developer builds.** One iteration is a day and 60 GB, behind 7 carried ORFS patches, generators for the register files, a floorplan planner and a stage-budgeted overnight harness. This week it also filled a 1.8 TB disk. It needs a build system and a campaign, not a `make`.
3. **What belongs in ORFS is what such a design extracts.** Every stopper this campaign found became a one-concern test that runs in seconds to minutes (table below): those are what a maintainer opens, edits and fixes at minimum cognitive load. The design that produced them stays with its build system, which regenerates the RTL from Chisel as fixes land upstream; a static RTL snapshot would be wrong within weeks and in the history forever.
4. **The frequency expectation has to go.** XiangShan's public target is 3 GHz at 7 nm, a 333 ps period. On asap7 through this flow, the best-placed block alone is at 0.9 to 1.9 ns after placement and repair, the synthesised core at 16.8 ns, and the core does not yet route.

### Where XSCore sits

where XSCore sits

XiangShan's repository ships XSCore (core, L1 caches, TLBs) inside an XSTile with a private 2 MB L2, tiles on a CHI network with a shared 32 MB LLC, and a NoC topology that supports one or two cores. This study places the innermost box only, in an integer-only configuration with a 512 KB L2 and 1 MB LLC outside it.

### The period ladder

period ladder

| measured on asap7 | minimum period | × target |
|---|---|---|
| XiangShan public target, 3 GHz at 7 nm | 333 ps | 1 |
| the four hardened blocks alone, placed and repaired at 800 ps | 0.92 to 1.9 ns | 2.8 to 5.7 |
| XSCore synthesised, before placement, at 800 ps (WNS −16.0 ns on the ROB glue) | 16.8 ns | 50 |
| XSCore after CTS, chain 4 (WNS −46.6 ns) | 47.4 ns | 142, an artefact: the blocks are abstracted at place, before their own CTS, so each abstract's clock pin carries the block's whole unbuffered clock net (MemBlock 145 pF) |

### One iteration (chain 4, 2026-09-22)

where the hours go

the die as run

| stage | wall | peak |
|---|---|---|
| synthesis, 48 partitions, slang | 2 h 13 | |
| floorplan, global placement | 19 + 38 min | 16 / 27 GB |
| repair_design | 33 min | 21 GB, 250,197 buffers |
| legaliser | 2 h 13 | 40 GB, all 1,400 negotiation iterations then the diamond search |
| CTS with `-no_insertion_delay` | 2 h 19 | 37 GB; the run before it inserted 49,736 delay buffers |
| route-0 gate at M9 | 22 min | 61 GB: 24 % of routing capacity used, worst edge 971 wires over, total congestion 81 M |
| global route, 5 iterations | killed at 3 h | inside the first maze iteration, as every knob arm of take 16 was |

### Stoppers, fixes, and the test each became

| found on the design | became |
|---|---|
| yosys `memory_dff` walks every module before asking which has a memory: 721 s per module | ORFS patch 0079, scoped passes, 64 s |
| slang `--keep-hierarchy` uniquifies memory names; blackboxing by definition | ORFS patch 0080 |
| a dead column of a placed array, removed by `eliminate_dead_logic`, stopped the placement script | patch 0081; `test/structured_netlist`, seconds, keeps the dead column |
| `rtl_macro_placer` refuses a fully fixed macro set | patch 0082; `test/planned_parent` |
| repair_design upsizes fixed array cells, then buffers an array's internal net into a dont_touch load (RSZ-3006), 3 h and 14 min | patch 0083; the three arrays at real size through CTS in `test/structured_netlist`, 10 min |
| CTS balances the parent to the blocks' unbuffered insertion delay: 49,736 delay buffers, 2.6 h of legaliser | `test/planned_parent` CTS delay-buffer test, minutes; `-no_insertion_delay` |
| FastRoute's usage guard (GRT-0228) in an 11 µm placement gap | `test/macro_select` two-block channel arm, 90 s |
| pin access (DRT-0073) fails on one pin per block after 9 h | 5 s reproducer on the miniature; cause open |
| the negotiation legaliser runs all 1,400 iterations before handing over | OpenROAD policy question; no test yet |
| one maze iteration takes hours on a 3,978² gcell grid | OpenROAD code question (take 16); no test yet |
| abstracts written before CTS export a 145 pF clock pin | miniature `abstract_stage` knob, planned |

### The ask

Take the extracted tests, not the design. Treat a request to check in a large design as a request to extract its concerns into tests of this size. The last three rows above are the open OpenROAD questions that extraction has not yet reached; they are where help is worth the most.

oharboe:
## What happens to this pull request

This branch is the campaign record, not a merge candidate. It carries 126 commits of how the XiangShan flow was got working, including every wrong turn worth remembering, and it stays open to be mined: the inventory in `ideas/xiangshan-timing.md` now has 23 entries, each tied to a measurement, and most of the reproducers, patches and feature requests that leave here start from one of them.

The baseline lands separately, as a new pull request built on `origin/main` with one commit per concern and no history of how we got here. Its goal is a single command:

```
bazelisk run //test/coremark_joule/designs/asap7/xiangshan:XSCore_plan_grt gui_grt
```

that reproduces today's result warts and all and opens it in the GUI, on any machine with 64 GB and the remote cache. Not a valid route: a fixed reference frame that interested parties can check out from `origin/main` and watch improve, one concern at a time. `test/coremark_joule/designs/asap7/xiangshan/README.md` on that branch says what the design is for, what is deliberately broken about it, and what to do with it instead of debugging against it.

The split, in the order the commits will land: the eight carried ORFS and OpenROAD patches, one each; the deployed-tree cross-stage guard; the generator's spare sites; the floorplan planner; the route bench; the three battery packages; the XiangShan generator and flow with the baseline arguments; the docs, the baseline record and the inventory.

What can split out from there as single-concern work, once each has a reproducer: to ORFS, the scoped memory passes (721 s to 64 s), memory detection by definition name under slang, dead cells in a placed netlist, no macro placer on a fully fixed set, and the resizer leaving placed-netlist cells alone; to OpenROAD, bounding and reporting the relocation search in `initialSnap`, a deterministic handover in the negotiation legaliser, hoisting the per-gcell layer lookup in the congestion writeback, and pin access on macro pins, which already has a five-second reproducer.

Last measurement on this branch, for the record: the floorplan raid halved route-0 congestion from 81.4 M to 41.8 M and the worst edge from 971 to 426, with the legaliser down from 2 h 13 to 39 min and clock tree synthesis from 2 h 19 to 57 min. Of what remains, 85 % is the parent's own wiring and nothing at all is in the channels and pockets the plan cleared. The next lever is not in the floorplan.

oharboe:
record of wip, cleaned up and merged to origin/main as a baseline

PR #1060, opened 2026-09-21
bazel-orfs: slang, always

oharboe:
bazel-orfs synthesises with slang. `SYNTH_HDL_FRONTEND` defaults to `slang` for
both public entry points that run synthesis — `orfs_flow()` and a standalone
`orfs_synth()` — and a design that sets another frontend is refused unless it
passes `yosys_frontend_reason = "<why>"` next to the setting. `SYNTH_USE_SYN=1`
bypasses yosys altogether and is exempt.

**Why a default and not a recommendation.** yosys's own Verilog frontend is
effectively deprecated for SystemVerilog and parses at a few MB/s. A design that
never said which frontend it wanted got yosys's, by ORFS's default, and paid for
it silently: XiangShan's flat core spent seven minutes of canonicalisation per
synthesis on it — and again after any input change — against one on slang.
Nobody had chosen the slow one; it is what happens when you say nothing.
`.claude/commands/prefer-slang.md` has been the standing recommendation for a
while. This makes it the default and makes the exception say why, in the BUILD
file, next to the setting.

**One policy, both entry points.** The first version of this landed in
`orfs_flow` only, which meant the same RTL synthesised through a different
frontend depending on which macro declared it. That is not hypothetical: it is
why `//test:discovered_modules_netlist_test` passed untouched while the two
`design()`-driven fixtures beside it — the same three-module design — moved to
slang and failed on a module name. The policy is one function now, and
`orfs_synth` pops `yosys_frontend_reason` because the rule has no such
attribute.

**Loud failure is the point.** A design whose SystemVerilog yosys accepted and
slang does not now fails at synthesis instead of running slowly. Two in reach
do:

- `riscv32i`'s `dmem.v` assigns `we_mem` and `mem_out` inside `always @*` while
  declaring them as nets — nine "cannot assign to a net within a procedural
  context" errors. A net cannot take a procedural assignment; yosys's frontend
  infers the variable anyway. Fixed at the two declarations and carried as
  `patches/0080-orfs-riscv32i-dmem-procedural-regs.patch`, so CI stays green.
- `tinyRocket` instantiates three `*_ext` memory wrapper modules its sources
  never define (its own `config.mk` says so). yosys blackboxes an undefined
  module implicitly; slang refuses it, and neither escape applies —
  `--ignore-unknown-modules` is gone from sv-elab, and `--blackboxed-module`
  marks a *defined* module. Left failing deliberately; it needs stub
  declarations ORFS-side. Its CI gate is `tags = ["manual"]`, so this is not
  visible to CI either way.

**Coverage.**

- `//test:slang_frontend_test` — the value the policy returns: the default, an
  explicit `slang`, a reasoned non-slang setting surviving, `SYNTH_USE_SYN`
  exempt in both directions.
- `//test:frontend_reason_synth` — a load-only `orfs_synth` carrying a reason.
  Swap its `pop` for a `get` and a bare query says `no such attribute
  'yosys_frontend_reason' in 'orfs_synth_rule' rule`.
- `test/synth_discovery_netlist_test.sh` now takes both frontends' spellings of
  a kept module (`adder` and `\adder$<top>.<instance>`), which is what ORFS
  already does for `SYNTH_KEEP_MODULES` in `synth.tcl`. Prefix collisions still
  fail: `adder2` does not answer for `adder`.
- The refusal itself has no test and cannot have one: it is a macro-level
  `fail()` at package load, so a target that must be refused cannot be declared
  anywhere. That is written down in the test file rather than left as a hole.

**Verified locally**, the three CI gates plus the suite:

- `bazelisk query '@orfs//flow/designs/...:*'` — every ORFS design package loads.
- `bazelisk build --nobuild` over every `@orfs` design `_route` target (67) and
  over `//...` — clean.
- `bazelisk test ... --build_tests_only` — 220/220. The last run executed exactly
  the two tests whose inputs changed (`discovered_modules_netlist_test`,
  `flattened_modules_netlist_test`, both now fed by a slang-synthesised netlist)
  and cached the rest. The netlist confirms it:
  `module \adder$kept_modules_top.u_adder`.

`TESTING.md`'s coverage row is updated: slang is no longer one package's setting.

Broken out of #1055. **Note for merge order:** that branch has since taken `0080`
for a patch of its own, so one of the two needs renumbering.


oharboe:
**Local suite: 216/216 pass.**

Three runs, each from the state named:

1. Policy alone — `//test:orfs_riscv32i_synth_build_test` failed to canonicalize.
   The flow log says only `ERROR: Compilation failed`; running `read_slang` by
   hand gave the nine "cannot assign to a net within a procedural context"
   errors in `dmem.v`. Fixed by ORFS patch 0080 above.
2. With 0080 — the ORFS patch list changed, so the archive refetched and 31
   tests re-ran from scratch. Two failed:
   `//test:dsl_hier_{discovery,pinned}_netlist_test`, both with
   `FAIL: no 'module adder'`. The hierarchy those fixtures exist to prove was
   intact — three modules in the merged netlist, one partition each — but slang
   uniquifies a kept module per instance, so the netlist holds
   `module \adder$kept_modules_top.u_adder`. The assertion now takes both
   spellings, which is what ORFS already does for `SYNTH_KEEP_MODULES`
   (`synth.tcl`). Prefix collisions still fail: `adder2` does not answer for
   `adder`, checked against a netlist holding only the mangled name.
3. Current tree — clean.

**Two gaps a reviewer should weigh, neither introduced by this branch but both
made reachable by it:**

- `yosys_frontend_reason` is an `orfs_flow` argument, and `design()` /
  `orfs_design` do not forward it. A design whose `config.mk` names a non-slang
  frontend would hit the `fail()` with no way to state a reason. Nothing in reach
  does today. The first one that needs the exception will need that plumbing.
- The default lands in `orfs_flow` only. A standalone `orfs_synth` (or other
  stage macro) still takes ORFS's default, so the same RTL can synthesise
  through a different frontend depending on which macro declared it. That is
  why `//test:discovered_modules_netlist_test` passed untouched while the two
  `design()`-driven fixtures next to it did not.

oharboe:
@maliberty slang always for bazel-orfs, yosys default verilog parser is broken for canonicalization and won't be fixed.

PR #1061, opened 2026-09-21
flow: user_stages is stripped in one place, and says why it stays in kwargs

oharboe:
Two leftovers from #1057, plus the coverage that was missing.

**The dead pop.** `update_kwargs.pop("user_stages", ...)` before `orfs_update` was
written when that rule was the only companion the sidecar reached.
`_strip_tool_kwargs` drops it for all three now, so the local pop does nothing
and its comment describes a bug fixed elsewhere.

**The warning on every run.** `bazelisk run //:fix_lint` printed

```
private/flow.bzl:364: function-docstring-args: Argument "user_stages" is
documented but doesn't exist in the function signature.
```

every time, for everyone. `orfs_flow` has no `user_stages` parameter on purpose:
the sidecar has to reach `_strip_tool_kwargs` still inside the `kwargs` dict,
because that function is the single place that decides which non-stage
companions must not see it. Lifting it into the signature would take it out of
that dict and move the decision back to each call site — the shape that failed.
The warning is disabled where that reason is now written down, rather than by
changing the plumbing to satisfy a linter.

**Coverage.** #1057 covers `orfs_arguments` and `orfs_macro` through a flow with
`user_stages` and a `mock_area`. It cannot cover `orfs_update`: that branch needs
a full flow with a `RULES_JSON` and *no* mock (`if "RULES_JSON" in test_args and
not mock_area`). So `//test:lb_32x128_update_user_stages_*` is declared beside
it, also load-only and never built. With the strip removed from
`_strip_tool_kwargs`, a bare `bazel query` on the package now names all three:

```
//test:lb_32x128_update_user_stages_update: no such attribute 'user_stages' in 'orfs_update' rule
//test:lb_32x128_mock_user_stages_mocked_generate_area: ... in 'orfs_arguments' rule
//test:lb_32x128_mock_user_stages_generate_abstract: ... in 'orfs_macro' rule
```

`TESTING.md` lists `orfs_update` under "NOT covered" — this is not a functional
test of the rule, but it is the first thing that instantiates it deliberately.


oharboe:
CI caught a mistake of mine in the fixture, now fixed.

The `orfs_update` fixture was meant to be load-only and was not. A flow with a
`RULES_JSON` generates a QoR test target, and `orfs_flow` clears the tags for
that one deliberately — `kwargs | {"tags": []} | test_kwargs` — so that a
design's QoR gate still runs when the design itself is `manual`. So the fixture
ran a full PnR in CI and checked it against `test/rules-base.json`, a rules file
written for another design:

```
[INFO] synth__design__instance__area__stdcell pass test: 5.832 <= 176.99
[INFO] constraints__clocks__count pass test: 2 == 2
[ERROR] Value not found for placeopt__design__instance__area.
```

`test_kwargs` is applied last, so it is the one way to say no; the fixture now
passes `test_kwargs = {"tags": ["manual"]}` and the reason is written beside it.

Locally: `bazelisk test ... --build_tests_only` finds 216 targets and all 216
pass, the same count as `main`, and the fixture's test appears nowhere in the
run. The `_update` target the fixture exists for is still declared, and removing
the strip from `_strip_tool_kwargs` still names all three companions.

Worth saying plainly: I verified this PR with targeted tests and the three lint
gates, not with the suite. CI ran the suite and I had not.

